PR #6, opened 2020-05-13
Updating README

bekozi:
* Added some further markdown formatting. Tried to use a uniform approach to syntax highlighting and adjusted some lists.
* Changed to wrapped approach for text blocks removing a lot of newlines.
* Remove some old text.
* Improved linking.
* Changed some header names to be more concise.

See https://github.com/esmf-org/esmf/tree/60f90c68773ac7d9b7516b554526587e839ddde7 to view in markdown format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,5 +54,3 @@ Demos, system tests, and use test cases, demonstrating how ESMF can be used in r
 
 Contributions from ESMF users are available at:
  * http://sourceforge.net/projects/esmfcontrib
-
-Please contact <esmf_support@ucar.edu> with any questions or problems.
--- a/README.md
+++ b/README.md
@@ -11,278 +11,34 @@ Hello and welcome to ESMF.
    example code, and details of the API (Application Programming Interface).
  * Please contact <esmf_support@ucar.edu> with any questions or problems.
 
-If you need to compile and link your code against an already-built ESMF library, go to the **APPLICATION** section below.
+## Downloading ESMF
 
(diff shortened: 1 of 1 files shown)


rsdunlapiv:
Some of the stuff in the README looks redundant with the user's guide.  Since there is no code in the user's guide (requiring build/testing) it might make sense to move the user's guide entirely to the GitHub wiki, or maybe to readthedocs, which allows for versioning.  Same with developers guide.  Of course, for the reference manual, we actually run the example code and the docs are intertwined with the API code, so I don't see how we could move that one out of the current process without significant effort.

bekozi:
We could move these sections to separate markdown files in the src/doc directory and link to them. What do you think? It would make the "splash page" text a bit more concise.

bekozi:
Got it. I think it is covered in the user's guide. Let me remove the sections and replace them with links and see how it goes.

bekozi:
I trimmed out a bunch of text by linking to the ESMF docs: https://github.com/esmf-org/esmf/blob/31c1cd26cfeedad20003d3458cee5baf788d6a07/README.md. Hopefully I didn't go too far.

We'll need to update the other add-on READMEs at some point.

bekozi:
I'm wondering if we should pull some text from here and put in the README: http://www.earthsystemmodeling.org/esmf_releases/public/last/ESMF_usrdoc/node2.html. We don't have any summary text.

rsdunlapiv:
Ben, I like the shorter README.  It is simpler and points to the relevant resources.

PR #17, opened 2020-08-28
Fix mesh dual creation on processors with no elements

rsdunlapiv:


rsdunlapiv:
Thanks @oehmke sounds good.  Denise is testing a build of the branch now - so let's at least wait to hear back from her before merging.

rsdunlapiv:
User reports successful test

PR #20, opened 2020-10-26
Per-language compile options with FindESMF.cmake

Chrismarsh:
If a custom Fortran compile options (`ESMF_F90COMPILEOPTS`) are  provided, these are saved in the esmf.mk file that is read by FindESMF.cmake. If a C or C++ program is compiled using FindESMF.cmake, these non-compatible options are passed to the C++ compiler causing possible errors. This change ensures per-language options are set in the target.

This change is per ticket 3615070.


theurich:
We need to add a C++ ESMF_HelloWorld.C and C ESMF_HelloWorld.c executable to the https://github.com/esmf-org/esmf-extended-tests/tree/develop/ESMF_HelloWorld_CMake. Then make sure our current FindESMF.cmake works for Fortran, C++, and C applications.
In particular, test the condition when a user sets explicit custom compile options, as per the comment by user.

theurich:
Support for C and C++ applications was supported, and examples were added to the extended-tests repo. See also follow-on discussion under https://github.com/esmf-org/esmf-support/issues/491.

PR #22, opened 2020-10-30
use std::inf and std::isnan for gcc5.4 compatibility

Chrismarsh:
Currently ESMF does not compile with gcc 5.4. This patch resolves that by ensuring the `isinf` and `isnan` functions are correctly located from the `std::` namespace.


rokuingh (approved):

rokuingh:
Tested by another intel user.

PR #27, opened 2021-03-23
Feature/tclune/abstract state item

tclune:
This request adds a new feature to ESMF that allows users to add objects of CLASS(AbstractStateItem) to an ESMF state.

In particular, users can add objects that are of concrete subclasses of AbstractStateItem and thereby add private state and type bound methods that can be used by other grid comps that are aware of the relevant subclass.    The idea is for sub communities (e.g., atmos. chemistry) to standardize an API and thereby share clean, lightweight interfaces for callbacks between say radiation and chemistry.

### DIFF
--- a/src/Superstructure/ESMFMod/src/ESMF.F90
+++ b/src/Superstructure/ESMFMod/src/ESMF.F90
@@ -92,6 +92,7 @@ module ESMF
     use ESMF_MeshMod
 
     use ESMF_StateTypesMod
+    use ESMF_AbstractStateItemMod
     use ESMF_StateVaMod
     use ESMF_StateMod
     use ESMF_StateReconcileMod
--- a/src/Superstructure/State/src/ESMF_AbstractStateItem.F90
+++ b/src/Superstructure/State/src/ESMF_AbstractStateItem.F90
@@ -37,17 +37,9 @@ module ESMF_AbstractStateItemMod
    type :: ESMF_AbstractStateItem
       sequence
       character(ESMF_MAXSTR) :: name
-      type(ESMF_AbstractItemWrapper) :: wrapper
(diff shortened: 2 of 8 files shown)


tclune:
Some comments from an earlier email exchange:

This branch clearly needs further work prior to incorporation into ESMF proper:
 
1.      I did not fully understand some of the macros being used in the source code and used brute force in a few places rather than following the usual ESMF conventions.
2.      I esp. do not yet understand the macros used in the unit tests, so my one test is really just a brute force procedure call inside the existing unit tests.    I.e., the test itself is meaningful but it’s not properly integrated.   More importantly it does not even pretend to exercise the various branches in the source code for optional arguments, already existing state items, …
3.      To avoid burning time writing the implicit interface in C, where my skills are very weak.   I instead exploited a hole in the Intel compiler to do this with Fortran’s BIND(C).    The resulting code is quite clear, but very non-standard in multiple respects.    Should be very straightforward for an expert to translate should that be desired.
4.      Names …    All procedure and type names should be decided by the core team.    And in particular, one of my types does not even have the ESMF prefix.  (Gasp!)
5.      Rather than just providing interfaces to exchange pointers to wrapper types ala ESMF_UserCompSetInternalState(),  I’ve also provided higher level interfaces that directly work with the abstract class directly.   Those higher level interfaces then construct wrappers accordingly for long term storage.    (And there might be a small memory leak there that might be unavoidable.   Needs more thought.)

And one more issue that I forgot in the email:     The abstract base class should have a deferred methods for serializing and deserializing objects.    This will facilitate use in concurrent execution, but will require further work inside ESMF to use these for transferring the objects.

tclune:
Closing this.   I think the ultimate users have moved on to other solutions.

PR #28, opened 2021-03-31
Add -DESMF_NO_F2018ASSUMEDTYPE to NAG and PGI on Darwin

mathomp4:
Testing by @tclune revealed a crash building with `nagfor` on macOS. Eventually, he figured out that the issue was it was trying to build F2018 code, which NAG can't (yet). 

So, until we know *what* versions of NAG can compile the code, I borrowed the:
```make
############################################################
# Currently NAG does not support the Fortran2018 assumed type feature
#
ESMF_F90COMPILECPPFLAGS += -DESMF_NO_F2018ASSUMEDTYPE
ESMF_CXXCOMPILECPPFLAGS += -DESMF_NO_F2018ASSUMEDTYPE
```
code from `Linux.nag` to add to `Darwin.nag`.  (NOTE: Not yet confirmed to work, @tclune is trying it now. So I'm making this a draft until we know.)

I also then added the same bits to the `Darwin.pgi` though I have no way of testing.

### DIFF
--- a/build_config/Darwin.nag.default/build_rules.mk
+++ b/build_config/Darwin.nag.default/build_rules.mk
@@ -116,6 +116,12 @@ ESMF_CLANGSTR := $(findstring clang, $(shell $(ESMF_CXXCOMPILER) --version))
 ESMF_F90COMPILEOPTS += -DESMF_NAG_UNIX_MODULE
 
 ############################################################
+# Currently NAG does not support the Fortran2018 assumed type feature
+#
+ESMF_F90COMPILECPPFLAGS += -DESMF_NO_F2018ASSUMEDTYPE
+ESMF_CXXCOMPILECPPFLAGS += -DESMF_NO_F2018ASSUMEDTYPE
+
+############################################################
 # nag currently does not support OpenMP
 #
 ESMF_OPENMP := OFF
--- a/build_config/Darwin.pgi.default/build_rules.mk
+++ b/build_config/Darwin.pgi.default/build_rules.mk
@@ -135,6 +135,12 @@ ESMF_F90COMPILECPPFLAGS += -DESMF_PGIVERSION_PATCH=$(ESMF_PGIVERSION_PATCH)
 ESMF_CXXCOMPILECPPFLAGS += -DESMF_PGIVERSION_PATCH=$(ESMF_PGIVERSION_PATCH)
 
 ############################################################
+# Currently PGI does not support the Fortran2018 assumed type feature
+#
+ESMF_F90COMPILECPPFLAGS += -DESMF_NO_F2018ASSUMEDTYPE
+ESMF_CXXCOMPILECPPFLAGS += -DESMF_NO_F2018ASSUMEDTYPE
+
+############################################################
 # Construct the ABISTRING
 #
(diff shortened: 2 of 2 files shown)


tclune:
I have verified that this change allows ESMF to build with NAG under Darwin.  I've also verified that MAPL can build with this version of ESMF and pass its own unit tests.  (Those do not exercise much of ESMF though.)

rsdunlapiv:
@mathomp4 @tclune would it be possible to see the output of "make all_tests" after the build?

rsdunlapiv:
@mathomp4 correct.  Perhaps you can attach as a file to this PR?

tclune:
This is in a new shell, so maybe something is inconsistent.
[test.log](https://github.com/esmf-org/esmf/files/6239460/test.log)

tclune:
And the info log.

[info.log](https://github.com/esmf-org/esmf/files/6239471/info.log)

rsdunlapiv:
Please rerun the info.log with all of the ESMF_* environment variables set.  The current info.log is showing gfortran as the compiler.  E.g., `ESMF_COMPILER=nag ESMF_BOPT=g ESMF_COMM=openmpi`

The test.log shows a failure due to not finding NetCDF.

If you have NetCDF in baselibs, can you please set:
`ESMF_NETCDF=/path/to/netcdf/nc-config`

Alternatively, you should be able to build and run the tests without NetCDF with ESMF_NETCDF not set at all.  Can you confirm that you do not have it set in your environment?

mathomp4:
@rsdunlapiv I think I have something. I had to run it multiple times to get things working, but it seems "consistent" now:
```
7791:Found 79 multi-processor examples, 78 passed and 1 failed.
8650:Found 48 multi-processor system tests, 48 passed and 0 failed.
10086:Found 3785 non-exhaustive multi-processor unit tests, 3417 passed and 368 failed.
```

[makeall_tests.nag-7.0_7036_openmpi-4.1.0.esmf.log](https://github.com/esmf-org/esmf/files/6243433/makeall_tests.nag-7.0_7036_openmpi-4.1.0.esmf.log)

rsdunlapiv:
@mathomp4 that is exactly what I am looking for.  For NAG, I guess prior to this you could not build at all, so this is at least getting us to a build of the lib and a lot of passing tests, but with some failures that need to be addressed at some point.  

For PGI, I assume you added:
```
ESMF_F90COMPILECPPFLAGS += -DESMF_NO_F2018ASSUMEDTYPE
ESMF_CXXCOMPILECPPFLAGS += -DESMF_NO_F2018ASSUMEDTYPE
```
based on the other build_config options for PGI?  This of course makes sense, but it is a shame that we don't have a way of testing this.  I have asked our local admins to install NAG and PGI on our own Mac server, but it might take a bit of time before those are set up.

I suspect that use of NAG and PGI on Darwin is relatively isolated.

@theurich I recommend we merge this in since it does at least move the needle forward on NAG/PGI support under Darwin.  When we have Catania set up with these compilers we will be able to debug some of these failing tests.  Thoughts?

theurich:
I agree that this looks correct, and I merged the PR. Also agree that it will be good to have testing for Darwin/nag and Darwin/pgi available to address the remaining test failures.

PR #41, opened 2022-03-18
Fix Resource File line limit

danrosen25:
TYPE: documentation fix

KEYWORDS: documentation, ESMF_ConfigGetLen, Resource File

SOURCE: Daniel Rosen, NCAR

DESCRIPTION OF CHANGES: Updates Resource File line limit in ESMF Reference Manual [46.1.2 Resource files](https://earthsystemmodeling.org/docs/release/latest/ESMF_refdoc/node6.html#SECTION06091200000000000000)
* see LSZ parameter in ESMF_Config.F90
* remove commented out LSZ code

ISSUES:
https://github.com/esmf-org/esmf-support/issues/155

### DIFF
--- a/src/Infrastructure/Config/doc/Config_usage.tex
+++ b/src/Infrastructure/Config/doc/Config_usage.tex
@@ -3,7 +3,7 @@
  \subsubsection{Resource files}
 
    A {\em Resource File (RF)} is a text file consisting of list of 
-   {\em label}-{\em value} pairs. There is a limit of 250 characters 
+   {\em label}-{\em value} pairs. There is a limit of 1024 characters 
    per line and the Resource File can contain a maximum of 200 records. 
    Each {\em label} should be followed by some data, the {\em value}. 
    An example Resource File follows.  It is the file used in the example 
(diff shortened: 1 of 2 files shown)


theurich (approved):
Looks good to be merged into develop.

danrosen25:
Hi @theurich 
I was thinking that we also add a check for line limitations during ConfigFileLoad. If a line is greater than max(1024,ESMF_MAXPATHLEN) then ConfigFileLoad should fail.  What do you think?

theurich:
@danrosen25 - I agree that would be good to have. Although, the implementation might be tricky. I figure that the truncation is done by the Fortran read... so to detect that a line was actually longer, you'd have to read into a longer buffer first?

theurich:
@danrosen25, I just came across this doc change PR again. Were you expecting me to merge this? If so, I probably dropped the ball.

danrosen25:
@theurich 
I'm comfortable with you merging it. I think the better fix needs to be implemented when there is more time.

theurich (approved):

PR #46, opened 2022-05-10
Fix FieldRead issue

jedwards4b:
Addresses issues in FieldRead and in ESMPy ugrid test.

### DIFF
--- a/src/Infrastructure/IO/src/ESMCI_PIO_Handler.C
+++ b/src/Infrastructure/IO/src/ESMCI_PIO_Handler.C
@@ -75,14 +75,14 @@ namespace ESMCI
   class PIO_IODescHandler {
   private:
     static std::vector<PIO_IODescHandler *> activePioIoDescriptors;
-    int ios;     // PIO IO system instance for this descriptor
-    int io_descriptor; // PIO IO descriptor
-    int nDims;                   // The number of dimensions for Array IO
-    int *dims;                   // The shape of the Array IO
-    int basepiotype;             // PIO version of Array data type
-    Array *array_p;              // The array matched to this descriptor
-    int arrayRank;               // The rank of array_p
-    int *arrayShape;             // The shape of array_p
+    int ios;                  // PIO IO system instance for this descriptor
+    int io_descriptor;        // PIO IO descriptor
+    int nDims;                // The number of dimensions for Array IO
+    int *dims;                // The shape of the Array IO
+    int basepiotype;          // PIO version of Array data type
+    Array *array_p;           // The array matched to this descriptor
+    int arrayRank;            // The rank of array_p
+    int *arrayShape;          // The shape of array_p
   public:
     PIO_IODescHandler(int iosArg, Array *arrayArg) {
       ios = iosArg;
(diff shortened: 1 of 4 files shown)


src/Infrastructure/IO/src/ESMCI_PIO_Handler.C
```diff
@@ -603,37 +603,30 @@ void PIO_Handler::arrayRead(
...
   localDE = 0;
   baseAddress = arr_p->getLocalarrayList()[localDE]->getBaseAddr();
   int arrlen = 1;
-  const int *counts = arr_p->getLocalarrayList()[localDE]->getCounts();
-  for (int i=0; i<narrDims; i++)
-      arrlen *= counts[i]; 
-
 #if defined(ESMF_NETCDF) || defined(ESMF_PNETCDF)
     int nDims;
-    if (((char *)NULL != name) && (strlen(name) > 0)) {
```
rsdunlapiv, line 622: @jedwards4b why was this part of the code removed?  I pushed a change just below this in the file to correctly set up varname.  Without it the varname is not being passed to PIOc_inq_varid below.  What's the right solution here?

jedwards4b: I think that it may have been accidental.  I don't recall removing it.

rsdunlapiv:
@theurich with these changes, all of the ESMPy issues are addressed in my local testing.  This is ready for your final review and merge.

theurich:
@rsdunlapiv I am running the branch through ESMF regression tests right now locally. If clean, I will merge the PR into develop. As for the code changes, I assume they are okay if tests okay also under ESMPy as you say. It's hard for me to say just looking at the code.

theurich:
@rsdunlapiv - Unfortunately I see this on the branch in local testing:

```
The following unit test files had failed unit tests:

FAIL: mpich/g: src/Infrastructure/Field/tests/ESMF_FieldIOUTest.F90


The following individual unit tests fail:

   FAIL: mpich/g:   Compare readin data to the existing data, ESMF_FieldIOUTest.F90, line 732:  Comparison failed
   FAIL: mpich/g:   Compare readin data to the existing data, ESMF_FieldIOUTest.F90, line 792:  Comparison failed

```

rsdunlapiv:
I was able to reproduce this error.  I think the issue might have to do with the "timeslice" argument to ESMF_FieldRead().

rsdunlapiv:
@theurich I have addressed the failure in ESMF_FieldIOUTest, which was due to trying to perform an ESMF_FieldRead() for multiple timeslices without an ungridded dimension on the Field to store the time dimension.  Actually, this test should have failed previously.  This test is now passing locally for me.  Can you please pull the branch and try again?

rsdunlapiv:
By the way, my local testing is still showing:
```
The following test harness unit tests fail, did not build, or did not execute:
FAIL: openmpi/g: ESMF_field_default_NP4UTest
```
Did you see this also @theurich?  Where do I go to see the "test harness" failures to try to diagnose this?

theurich:
@rsdunlapiv - the regression tests under this branch now are 100% clean under my local testing. I think this problem is resolved, and I am going to proceed with merging the PR.

With respect to the "test harness" failures, I suspect that you may not have built all of the unit test executables, and instead only built locally under Field/tests. When you do this, the ESMF_TestHarnessUTest executable is not available, but the Field test harness configuration lives under Field/tests. This is why the FAIL. If you go and build the ESMF_TestHarnessUTest, and then come back under Field/tests, and run again, it should all be clean.

PR #47, opened 2022-05-16
Add explicit casts from PIO_offset to int

billsacks:
This is needed to avoid compilation errors like this (using clang on my Mac):

/Users/sacks/esmf/esmf0/src/Infrastructure/Mesh/src/ESMCI_ESMFMesh_Util.C:588:21: error: non-constant-expression cannot be narrowed from type 'MPI_Offset' (aka 'long long') to 'int' in initializer list [-Wc++11-narrowing]
  int gdimlen2D[2]={elementCount,maxNodePElement};

I'm not positive this change is correct, but I do see similar casts elsewhere - such as:

https://github.com/esmf-org/esmf/blob/de129fc120c40a5e5d3ce85081d5883dce53748e/src/Infrastructure/Mesh/src/ESMCI_ESMFMesh_Util.C#L269

### DIFF
--- a/src/Infrastructure/Mesh/src/ESMCI_ESMFMesh_Util.C
+++ b/src/Infrastructure/Mesh/src/ESMCI_ESMFMesh_Util.C
@@ -585,7 +585,7 @@ void get_elemConn_info_2Dvar_from_ESMFMesh_file(int pioSystemDesc, int pioFileDe
 
   // Init elementConn decomp
   int ec_iodesc;
-  int gdimlen2D[2]={elementCount,maxNodePElement};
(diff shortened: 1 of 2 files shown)


theurich:
@billsacks I think these might be the same casts that are currently under https://github.com/esmf-org/esmf/tree/fix/cce. Which probably does make sense because CCE is using CLANG. Could you see if the fix/cce includes all the casts that you need for your test environment?

theurich:
Sounds good @billsacks, I am still working on fix/cce, but mostly on CCE related configurations now. Also it is waiting for testing by @jedwards4b under the CCE test setup on Cheyenne. I hope it will be merged into develop very soon.

PR #48, opened 2022-06-07
Install FindESMF.cmake to prefix/lib/cmake/esmf

mathomp4:
TL;DR: This PR installs the `FindESMF.cmake` file when running `make install`

---

One of my goals with MAPL (and GEOS) is to eventually be able to build using [Spack](https://github.com/spack/spack) as my library provider instead of Baselibs. However, testing today with building the latest MAPL with Spack led to an issue. Namely, recently in GEOS CMake, we've moved to using `find_package(ESMF MODULE REQUIRED)` and then appending to our `CMAKE_MODULE_PATH` where that file is in our installation of ESMF in Baselibs. We used to carry around our own `FindESMF.cmake`, but then I realized @danrosen25 and @theurich do change it every so often, so the "right" thing is to just use what ESMF itself provides.

But, the Spack testing revealed this all works only because in the Baselibs installation, I [*explicitly* copy the file](https://github.com/GEOS-ESM/ESMA-Baselibs/blob/594adb60cbbe86b508f6689e0f6fbf4cb0f9dd1c/esmf_rules.mk#L158) into our install directory:

```make
	@cp -pr $(ESMF_DIR)/cmake/*    $(ESMF_INSTALL_HEADERDIR)
```

because the ESMF `make install` does not do this. 

So, in this PR, I do the "simple" thing and add two lines to the `makefile` so that the `FindESMF.cmake` file is copied into `$(ESMF_INSTALL_LIBDIR_ABSPATH)/cmake/esmf`. 

Note: this is different than how I am currently doing it in Baselibs (see above), but I just sort of winged that long ago, but `lib/cmake/esmf` seems to be a more "canonical" path for a CMake package installations. Now, most packages built with CMake install for Config Mode and, [per this page](https://cmake.org/cmake/help/latest/command/find_package.html#config-mode-search-procedure), the first path CMake looks on Unix for Config mode installs is:
```
<prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/                 (U)
```
So, `lib/cmake/esmf/` seems a good place even for a Module-mode install.

Now, the question for you is: is this how you'd like to do it? Maybe you'd like to make an `ESMF_INSTALL_CMAKEDIR` or some such override variable. I dunno. And maybe you'd also like to copy the file into, say, `lib/libO/Darwin.nag.64.openmpi.default/cmake/esmf` as well? Not sure.

I'd love for this to be in v8.3.0 but I figure that train has left the station. 😄  For now, I have a "hack" in how I'm building MAPL in Spack testing that seems to work where I just grab the file blob from GitHub and shove it into where Spack is building MAPL. But if `make install` installed it, then I think I could work with/bother @climbfuji (who maintains the [esmf Spack package](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/esmf/package.py)) to figure out how to get Spack to add the `prefix/lib/cmake/esmf` to the `CMAKE_MODULE_PATH` Spack can generate.

### DIFF
--- a/makefile
+++ b/makefile
@@ -699,6 +699,8 @@ install:
 	cp -f $(ESMF_MODDIR)/*.mod $(ESMF_INSTALL_MODDIR_ABSPATH)
 	mkdir -p $(ESMF_INSTALL_LIBDIR_ABSPATH)
 	cp -f $(ESMF_LIBDIR)/libesmf*.* $(ESMF_INSTALL_LIBDIR_ABSPATH)
+	mkdir -p $(ESMF_INSTALL_LIBDIR_ABSPATH)/cmake/esmf
+	cp -f $(ESMF_DIR)/cmake/* $(ESMF_INSTALL_LIBDIR_ABSPATH)/cmake/esmf
 
 ifeq ($(ESMF_PIO),internal)
 	cp -f $(ESMF_LIBDIR)/libpioc.* $(ESMF_INSTALL_LIBDIR_ABSPATH)


mathomp4:
> @danrosen25 - I agree. Also let me include @rsdunlapiv here.
> 
> Personally I keep going back to the CMAKE documentation itself on this topic, specifically to https://cmake.org/cmake/help/latest/command/find_package.html
> 
> The second paragraph is pretty clear about the intent of `Find<PackageName>.cmake` not being provided by the package itself. That is why we make such a point about it in ESMF: we provide FindESMF.cmake as a template to be included in the layer that wants to use ESMF.
> 
> The reasoning that the package itself cannot be the provider of `Find<PackageName>.cmake` also makes pretty good sense to me. If the package itself were to provide `Find<PackageName>.cmake`, then you need to already know where the package is located in order to use `Find<PackageName>.cmake`.... but that defeats the whole purpose of `Find<PackageName>.cmake`, doesn't it?

@theurich Huh. Interesting. I guess I never read that bit of the `find_package` page, I just went further down. So maybe this is a "dead" PR. I'll consult with @tclune

climbfuji:
Note. We have FindESMF.cmake in noaa-emc/CMakeModules and in jcsda/jedi-cmake, nearly identical versions. Looking towards converging on one.

We use that in all our software packages and install it with our spack stack solution. This spack stack will be released first time end of this month (June 2022) and will support all JCSDA JEDI and UFS applications.

mathomp4:
> Maybe @mathomp4 would consider retiring Baselibs and jumping on to a community-wide spack dependency management solution that would be shared among the agencies?

@rsdunlapiv I am hoping to move that way, but it might be a while until we can. Need to feed up all of Tom's packages to Spack, then I think I still might have a couple outstanding issues with Spack as well (mainly to do with the Intel compilers; oneAPI did a number to my testing).

mathomp4:
Close for more thought

mathomp4:
One update from the [CMake Discourse](https://discourse.cmake.org/t/advice-for-maintaining-module-mode-findpackage-cmake-files/5818/2?u=mathomp4):

> I think the way to go is rather convince ESMF to provide config files instead of find module and install that. One does not need CMake build system to generate those (as, for example, boost proves)

I'm guessing this is the file he is referring to:

https://github.com/boostorg/boost_install/blob/85e1406a570cb8aa13c8b1f885f7cc2a9982c783/BoostConfig.cmake

That is pretty complex, but Boost might be the _ne plus ultra_ of complex libraries (159 submodules!!).  But I did speak with @tclune and he thinks gFTL-shared might provide a good "template" to start from. 

Now, gFTL-shared is built with CMake, so that's essentially using CMake to generate the output from a template. The "fun" with ESMF would be trying to figure out how to use GNU Make to fill out a template. We could do the "have CMake pull from esmf.mk" like the Config file does now, but it would be nice if at `make install` time, GNU Make would fill it out and then the installed file would be "static" and not depend on anything. I guess there is sed and/or m4...

I'll try and ponder this, but it is a possible way forward.

PR #53, opened 2022-07-19
Fix some documentation

billsacks:
It looks like some parts of the documentation for
ESMF_GridCreateMosaicReg were copied from the documentation for
ESMF_GridCreateCubedSphereReg. Thus, the former had some references to a
fixed tile count of 6 when in fact that routine supports an arbitrary
tile count.

In addition, fix some minor typos.


billsacks:
@rsdunlapiv I believe this is a straightforward fix to the documentation. However, given my newness to ESMF, I wanted to let someone review it.

I have maintained the line breaks where they were before to facilitate review, but if you prefer I can rewrap the paragraph before merging to avoid long lines.

rsdunlapiv (approved):

rsdunlapiv:
@billsacks looks good so I will merge.  It looks like I need to fix CircleCI to pull from forks.  That's why the docs failed to build on your PR.

PR #54, opened 2022-07-25
Enable I/O for multi-tile arrays

billsacks:
Changes to enable I/O for multi-tile arrays. The overall approach is to loop over tiles, creating a separate I/O descriptor (decomposition) for each tile, and writing / reading each tile to / from its own file.


billsacks:
Changes so far are:

(1) Introducing a new function on the distgrid class (getTilePLocalDe) which gives the tile index for a given local DE. This encapsulates the previous multi-stage process that involved translating the local DE to a global value then using that to index into an array. In addition to using this new function in the multi-tile I/O code, I have applied this new function in a few other parts of the framework that previously were using this multi-stage process.

(2) Fixing some aspects of the PIO decomposition calculation so that this calculation works for a given tile of a multi-tile array.

(3) Fixing part of the PIO write code so that it works for a single tile of a multi-tile array.

There is still a fair amount to be done, particularly extending the code to use loops over tiles and store various information in arrays of length num_tiles, and breaking the input / output into multiple files.

.gitattributes
```diff
@@ -1,3 +1,5 @@
 src/doc		export-ignore
 .gitattributes	export-ignore
 .gitignore	export-ignore
+
+*.C linguist-language=C++
```
billsacks, line 5: GitHub seemed to be syntax highlighting ESMF .C source files as C rather than C++. This seems to fix it.

src/Infrastructure/DistGrid/src/ESMCI_DistGrid.C
```diff
@@ -4962,6 +4962,52 @@ ESMC_I8 DistGrid::getElementCountPDe(
...
 //-----------------------------------------------------------------------------
 
+//-----------------------------------------------------------------------------
+#undef  ESMC_METHOD
+#define ESMC_METHOD "ESMCI::DistGrid::getTilePLocalDe()"
+//BOPI
+// !IROUTINE:  ESMCI::DistGrid::getTilePLocalDe
+//
+// !INTERFACE:
+int DistGrid::getTilePLocalDe(
```
billsacks, line 4972: Note this new function. This allows replacing a multi-step process (converting local DE to global and then using that to index into an array) with a single function call. I didn't replace all possible uses: I focused on uses of the old pattern where we wouldn't otherwise need to do the local to global DE conversion, because that's where the simplification is greatest.

src/Infrastructure/DistGrid/src/ESMCI_DistGrid.C
```diff
@@ -5099,8 +5145,9 @@ template<typename T> int DistGrid::tGetSequenceIndexLocalDe(
       tArbSeqIndexListPCollPLocalDe[0][localDe][linExclusiveIndex]);
   }else{
     // determine the sequentialized index by construction of default tile rule
-    const int *localDeToDeMap = delayout->getLocalDeToDeMap();
-    int tile = tileListPDe[localDeToDeMap[localDe]];  // tiles are basis 1 !!!!
+    int tile = getTilePLocalDe(localDe, &localrc);
+    if (ESMC_LogDefault.MsgFoundError(localrc, ESMCI_ERR_PASSTHRU,
+      ESMC_CONTEXT, &rc)) return rc;
```
billsacks, line 5150: I would especially welcome a close review of my new MsgFoundError error handling calls, here and elsewhere. I have tried to replicate similar error handling elsewhere, with particular attention given to the use of localrc vs. rc (which appears to be a bit inconsistent between different parts of the code). But it's possible that I'm misusing this error handling in some places.

Also note lack of error handling in some situations like in `GridIter::setDEBnds` (in ESMCI_Grid.C), which didn't seem to be set up to accommodate error handling.

src/Infrastructure/IO/src/ESMCI_PIO_Handler.C
```diff
@@ -808,13 +812,24 @@ void PIO_Handler::arrayWrite(
...
+  if (tileOfThisDe == tile) {
+    baseAddress = arr_p->getLocalarrayList()[localDE]->getBaseAddr();
+    PRINTMSG("baseAddress = 0x" << (void *)baseAddress);
+    // arrlen = arr_p->getLocalarrayList()[localDE]->getByteCount();
+    arrlen = 1;
+    const int *counts = arr_p->getLocalarrayList()[localDE]->getCounts();
+    for (int i=0; i<narrDims; i++)
+      arrlen *= counts[i];
+  } else {
+    baseAddress = NULL;
```
billsacks, line 830: Passing a NULL to PIO from procs that don't own this tile feels like it will prevent inadvertently trying to read from / write to portions of memory that shouldn't be read from / written to for this call. However, I'm not 100% sure that this passing of NULL is the right thing to do in this situation. I'd like to confirm that, possibly with Jim Edwards.

src/Infrastructure/IO/src/ESMCI_PIO_Handler.C
```diff
@@ -2021,28 +2050,46 @@ int PIO_IODescHandler::constructPioDecomp(
     handle->arrayShape = (int *)NULL;
   }
   handle->arrayShape = new int[handle->arrayRank];
-  for (int i = 0; i < handle->arrayRank; ++i) {
-    handle->arrayShape[i] = (totalUBound[(tile * handle->arrayRank) + i] -
-                             totalLBound[(tile * handle->arrayRank) + i] +
-                             1);
```
billsacks, line 2027: Note that I changed the calculation of arrayShape here: I'm pretty sure it is incorrect to have tile in here: this is documented as having dimension [redDimCount*ssiLocalDeCount]. So in principle we should have the DE number in here, but we can leave that out as long as we assume no more than one DE per PE.

billsacks:
Test program and data I'm using for manual testing of this so far: https://github.com/billsacks/esmfprojects-multi_tile_io . I will convert this to a unit test in the ESMF repo at some point.

billsacks:
One other question I have about pre-existing code: In `PIO_IODescHandler::constructPioDecomp`, I noticed that some functions get values specific to the SSI (single system image) whereas others seem not to. Is this correct? e.g., `arr_p->getTotalLBound()` and `arr_p->getTotalUBound()` get variables whose size is based on `ssiLocalDeCount`; but the functions on the DELayout like `getLocalDeCount` and `getLocalDeToDeMap` do not appear to be related to the SSI, at least not explicitly. Is this difference a problem?

theurich:
Bill,

Not sure about the code in question, but from what you are saying it is 
probably okay. The way that DE sharing across SSI's works is that the 
shared DEs are _following_ the owned DEs in the arrays on each PET. So 
on a specific PET, these arrays do include elements for the shared DEs, 
but if the code only access the PET local DEs, at the beginning of these 
arrays, all should be good. It sounds like that is what is going on.

-Gerhard

billsacks:
> Not sure about the code in question, but from what you are saying it is
probably okay. The way that DE sharing across SSI's works is that the
shared DEs are _following_ the owned DEs in the arrays on each PET. So
on a specific PET, these arrays do include elements for the shared DEs,
but if the code only access the PET local DEs, at the beginning of these
arrays, all should be good. It sounds like that is what is going on.

Thanks @theurich - though I must admit that I'm not completely following this. It sounds like this is okay here, but I'd like to talk to you or someone else at some point to try to understand this better for the sake of my long-term understanding of the ESMF architecture - since this strikes me as something where a lack of understanding might lead to subtle coding errors (or maybe not).

billsacks:
By changing the test program (https://github.com/billsacks/esmfprojects-multi_tile_io) to write the actual values of latitude and longitude, I was able to compare the current tile-specific output with the true values extracted from the grid files. By changing two hard-coded instances of `tile = 1` in `ESMCI_PIO_Handler.C` to each of the possible values from 1 - 6, I have verified that the tile-specific write is indeed writing out the correct values for each tile now.

(I did this testing on 48bec3df28 (changing two instances of `tile = 1` to each possible value from 1 to 6), and billsacks/esmfprojects-multi_tile_io@60da335 (but with the singletile write commented out)).

billsacks:
Among other recent changes, I just pushed a commit (7d8086b) that allows separate IO descriptors for each tile. Related to this, @rsdunlapiv  asked me to look at whether there will be a constraint of all arrays having the same dimensions for each tile... and if not, how that's going to be handled. With the way I have set up the different IO descriptors (via the PIO_IODescHandler class), I don't think this will be an issue: we'll have a separate IO descriptor for each array and tile, so the IO descriptors for two different arrays will remain completely independent. (With the caveat that I may be missing something since I still don't know all of the relevant code.)

billsacks:
I have done the following testing so far on c764928372711b54776023c9eec6cffe954038ad together with billsacks/esmfprojects-multi_tile_io@d675089 (note that I referenced the wrong commit in that test repo in my commit message for c764928372711b54776023c9eec6cffe954038ad):
- `make all_tests` on my Mac with gcc / gfortran: all tests pass except for 6 unit tests that fail on develop as well. (I also have testing underway on cheyenne with intel; I'll comment here if that testing shows any issues)
- With my test program:
  - Confirming that the output fields match the expectations, by comparing them against the lat and lon values in the original grid files, which they have been set up to reproduce
  - Confirming that all 3 decompositions of the multi-tile arrays give identical output results
  - Eyeball comparison of x vs x_copy and y vs y_copy for all 6 tiles, seeing that these "copy" variables are identical (which helps confirm, for example, that it is working to pick up an already-computed decomposition)

billsacks:
'make all_tests' passes on cheyenne with the intel compiler.

As discussed on this morning's call, I am about to update this branch to the latest version of develop.

billsacks:
@uturuncoglu note that I have updated to latest develop and pushed to this branch. So please pull down the new version if you haven't already. I have rerun both my manual testing of the new code as well as 'make all_tests' on my Mac and both look good with this updated version.

billsacks:
I have updated the API documentation for fileName. I would be happy for someone to look my changes over to make sure they look good (https://github.com/esmf-org/esmf/pull/54/commits/0c49e34a6ee6252a3e50ed4acea57b466f857053). Note that I also documented the limitation that, for FieldBundleRead/Write and ArrayBundleRead/Write, all Fields / Arrays must have the same number of tiles... my impression is that this is a non-issue for FieldRead/Write and ArrayRead/Write, but let me know if I'm wrong about that and those subroutines actually have the potential to work on multiple fields / arrays (in which case there would be the same limitation).

I also made an unrelated documentation fix regarding valid values of ESMF_PIO (https://github.com/esmf-org/esmf/pull/54/commits/959029a2d976dfc423dbef17b2d5b4a946b935ab).

billsacks:
@rsdunlapiv @theurich and others: I have just pushed a commit that adds unit tests. So far I am testing FieldBundleWrite and FieldWrite. Next I will add tests of the associated Reads and get them working. Then I'll add tests of ArrayBundle and Array (I expect those to work right now, but I wanted to move on to getting the Reads working before filling out the remainder of this unit test program).

A few notes about this that I welcome feedback on:
1. I have tentatively named the test file to be specific to multi-tile IO, in contrast to my originally-proposed name (something like ESMF_IO_HighLevelUTest). This is because I felt that it would be confusing to have unrelated tests grouped together in one file, particularly given that the standard pattern seems to be to have all tests in the main program or internal subroutines of that main program, so having different groups of unrelated tests all together could get hard to navigate and hard to know which variables pertain to which tests. And this way we can use 8 procs just for these multitile tests.
2. When I started writing these, I thought it would make sense to bail out early if prerequisite steps such as the createFields subroutine failed, since if those fail, then the remaining unit tests will fail in probably cryptic ways. But from spot-checking a few other files, I don't see that done elsewhere: it seems like it's typical for the entire set of unit tests to try to run even if there is a failure in some prerequisite. So I've left it like that here, too - i.e., *not* bailing out even if prerequisites fail. I'm open to feedback on this.
3. I have made some of the tests exhaustive-only. My rationale was to get pretty much full code coverage with the non-exhaustive tests, then make the mostly-redundant tests exhaustive-only. Is that in line with what's typically done, or should I make more or fewer of my tests exhaustive-only?

theurich:
Hi Bill, see my responses below...

On 8/26/22 15:01, Bill Sacks wrote:
>
> @rsdunlapiv <https://github.com/rsdunlapiv> @theurich 
> <https://github.com/theurich> and others: I have just pushed a commit 
> that adds unit tests. So far I am testing FieldBundleWrite and 
> FieldWrite. Next I will add tests of the associated Reads and get them 
> working. Then I'll add tests of ArrayBundle and Array (I expect those 
> to work right now, but I wanted to move on to getting the Reads 
> working before filling out the remainder of this unit test program).
>
> A few notes about this that I welcome feedback on:
>
>  1. I have tentatively named the test file to be specific to
>     multi-tile IO, in contrast to my originally-proposed name
>     (something like ESMF_IO_HighLevelUTest). This is because I felt
>     that it would be confusing to have unrelated tests grouped
>     together in one file, particularly given that the standard pattern
>     seems to be to have all tests in the main program or internal
>     subroutines of that main program, so having different groups of
>     unrelated tests all together could get hard to navigate and hard
>     to know which variables pertain to which tests. And this way we
>     can use 8 procs just for these multitile tests.
>
That does make a lot of sense to me. Honestly in the discussion 
yesterday I was thinking of "HighLevel" more as a generic placeholder 
e.g. for Multitile, or similar. Glad this is how it went in.
>
>  1. When I started writing these, I thought it would make sense to
>     bail out early if prerequisite steps such as the createFields
>     subroutine failed, since if those fail, then the remaining unit
>     tests will fail in probably cryptic ways. But from spot-checking a
>     few other files, I don't see that done elsewhere: it seems like
>     it's typical for the entire set of unit tests to try to run even
>     if there is a failure in some prerequisite. So I've left it like
>     that here, too - i.e., /not/ bailing out even if prerequisites
>     fail. I'm open to feedback on this.
>
In some cases you will see that very basic calls are error checked and 
terminated via

if (rc /= ESMF_SUCCESS) call ESMF_Finalize(endflag=ESMF_END_ABORT)

These are calls that are not really unit checked in a particular file, 
but just supportive ESMF calls that are made. Not expected to fail, but 
if they do, there is an immediate abort.

As for sequences of actual unit testing via the ESMF_Test() method, it 
is correct that we will just keep going. The philosophy is that any 
follow-on calls need to also be robust to be able to handle problems in 
caused by earlier calls. In fact we sometimes even purposefully trigger 
error conditions (and test that they are caught). In the end of the day, 
this approach makes it much easier to write and maintain these UTest 
files. When something early on fails that is a prerequisite, you get 
tons of failures... that is okay.. and needs to be fixed.

>  1. I have made some of the tests exhaustive-only. My rationale was to
>     get pretty much full code coverage with the non-exhaustive tests,
>     then make the mostly-redundant tests exhaustive-only. Is that in
>     line with what's typically done, or should I make more or fewer of
>     my tests exhaustive-only?
>
There is quite a bit of grey area, between exhaustive and 
non-exhaustive. Testing basic complete functionality for non-exhaustive, 
and then going into finer granularity, even with some redundant testing 
for exhaustive, as you say, is a good working guideline.

-Gerhard

billsacks:
Thanks a lot for the input, Gerhard. This makes sense.

One other general note about how I have set this up so far: I introduced internal subroutines to group together various setup / creation steps - e.g., creating the Fields and FieldBundle that I use in the actual tests - and then treat the execution of each of those setup subroutines as a single unit test. Alternatives would be to treat nearly every ESMF call as its own unit test (this seems to be what's done in ESMF_FieldBundleIOUTest.F90, which is the first example I looked at), or to not call this initial setup a unit test at all. I felt like what I did was a kind of compromise between those two alternatives, but please let me know if there is a preference in this respect.

rsdunlapiv:
@billsacks the approach of treating the setup calls themselves as a unit test makes sense to me and seems the right compromise between having a very large number of unit tests and assuming that the setup will succeed.

src/Infrastructure/IO/src/ESMCI_PIO_Handler.C
rsdunlapiv: @jedwards4b can you please confirm Bill's logic on setting baseAddress to NULL when procs do not own a tile to be read/written?

src/Infrastructure/IO/src/ESMCI_IO_Handler.C
```diff
@@ -361,6 +366,101 @@ void IO_Handler::finalize (
...
+//-----------------------------------------------------------------------------
+  // initialize return code; assume routine not implemented
+  int localrc = ESMF_RC_NOT_IMPL;         // local return code
+  if (rc != NULL) {
+    *rc = ESMF_RC_NOT_IMPL;               // final return code
+  }
+
+  PRINTPOS;
+
+  for (int tile = 1; tile <= ntiles; ++tile) {
```
rsdunlapiv, line 452: @billsacks for this loop (and similarly for reads) are we getting any concurrency on the write when tiles are on completely disjoint sets of procs?

billsacks: > for this loop (and similarly for reads) are we getting any concurrency on the write when tiles are on completely disjoint sets of procs?

Unfortunately, no. Getting concurrency would require a more substantial rework of the PIO interface so that we have separate PIO instances for each tile. From our earlier discussions, we decided not to do that for the first iteration.

jedwards4b (approved):

PR #72, opened 2022-09-19
update to pio2_5_9

jedwards4b:
Update pio external to pio2.5.9


src/Infrastructure/IO/PIO/ParallelIO/CMakeLists.txt
```diff
@@ -4,13 +4,13 @@
 
 # Jim Edwards
 
-cmake_minimum_required (VERSION 3.5.2)
+cmake_minimum_required (VERSION 3.7)
```
theurich, line 7: @jedwards4b - upping the cmake_minimum_requiremed is an issue for ESMF. Is it possible to stay with CMake 3.5.2?

theurich (changes requested):
We are seeing build failures due to the change in CMake minimum version requirement.

src/Infrastructure/IO/PIO/ParallelIO/CMakeLists.txt
jedwards4b: Requiring cmake 3.7 which was released 12/2018 is not exactly a cutting edge software requirement. 
I'm looking back at my notes and can't find exactly the reason this minimum was changed.  It had something to
do with supporting spack.   I just tried a standalone build with 3.5.2 and it appears to work fine.

PR #77, opened 2022-09-28
Correct esmpy documentation for extrap method

mdunphy:
Hi, it looks like the paragraphs for `NEAREST_STOD` and `NEAREST_IDAVG` are switched here, so this commit switches them back


rokuingh:
@mdunphy, good catch, thanks!

PR #79, opened 2022-10-03
Fixes to get Darwin.gfortranclang combination working; add Linux ARM and LLVM compilers

### LINKED ISSUE #60: Support clang compiler on Darwin (Mac OS X)
Many tests fail or hang with the clang compiler (with ESMF_COMPILER=gfortranclang), at least on my Mac.

The build setup was like this:
```
export ESMF_DIR=/Users/sacks/esmf_builds/esmf8.3.0g
export ESMF_INSTALL_PREFIX=$ESMF_DIR/install_dir
export ESMF_NETCDF=split
export ESMF_NETCDF_INCLUDE=/opt/homebrew/include
export ESMF_NETCDF_LIBPATH=/opt/homebrew/lib
export ESMF_COMM=openmpi
export ESMF_COMPILER=gfortranclang
export ESMF_BOPT=g
```

Some tests pass, but it seems like many / all tests that involve I/O are either crashing or hanging. From spot-checking two tests (ESMF_RouteHandleAdvancedUTest and ESMF_FieldBundleIOUTest) they die in the same place. The error message printed is:

Program received signal SIGTRAP: Trace/breakpoint trap

which sounds as if there is an explicit breakpoint being set somewhere?

Running in a debugger, I see that the error occurs while destroying an IO_Handler object, and specifically it looks like it's in the destructor for a moab Tree object, though I haven't seen where that Tree object comes from. Print-based debugging points to the line "delete (*ioclass);" in ESMCI_IO_Handler.C (but I can't see where the error occurs at a lower level than that). Here is the stack trace from lldb:

```
thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BREAKPOINT (code=1, subcode=0x102d721dc)
    frame #0: 0x0000000102d721dc libesmf.dylib`moab::Tree::~Tree()
libesmf.dylib`moab::Tree::~Tree:
->  0x102d721dc <+0>: brk    #0x1

libesmf.dylib`std::__1::deque<ESMCI::VMK::ackElement, std::__1::allocator<ESMCI::VMK::ackElement> >::__add_back_capacity(unsigned long) (.cold.1):
    0x102d721e0 <+0>: stp    x20, x19, [sp, #-0x20]!
    0x102d721e4 <+4>: stp    x29, x30, [sp, #0x10]
    0x102d721e8 <+8>: add    x29, sp, #0x10            ; =0x10
Target 0: (ESMF_RouteHandleAdvancedUTest) stopped.
(lldb) bt
thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BREAKPOINT (code=1, subcode=0x102d721dc)
  * frame #0: 0x0000000102d721dc libesmf.dylib`moab::Tree::~Tree()
    frame #1: 0x00000001020de7f4 libesmf.dylib`ESMCI::IO_Handler::destroy(ESMCI::IO_Handler**) + 504
    frame #2: 0x00000001020d3abc libesmf.dylib`ESMCI::IO::destruct() + 120
    frame #3: 0x00000001020d71dc libesmf.dylib`ESMCI::IO::~IO() + 20
    frame #4: 0x00000001020d4254 libesmf.dylib`ESMCI::IO::~IO() + 12
    frame #5: 0x00000001020d4088 libesmf.dylib`ESMCI::IO::destroy(ESMCI::IO**) + 124
    frame #6: 0x0000000101eabaac libesmf.dylib`ESMCI::Array::write(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, bool*, ESMC_FileStatus_Flag*, int*, ESMC_IOFmt_Flag*) + 844
    frame #7: 0x00000001022fe7d8 libesmf.dylib`ESMCI::RouteHandle::print() const + 2200
    frame #8: 0x0000000102301ddc libesmf.dylib`c_esmc_routehandleprint_ + 116
    frame #9: 0x000000010288f204 libesmf.dylib`__esmf_rhandlemod_MOD_esmf_routehandleprint + 472
    frame #10: 0x0000000100005a84 ESMF_RouteHandleAdvancedUTest`MAIN__ + 6828
    frame #11: 0x0000000100007124 ESMF_RouteHandleAdvancedUTest`main + 32
    frame #12: 0x000000010002508c dyld`start + 520
```

Most or all of the `SIGTRAP` errors I see seem to be in tests involving I/O based on their name. There is also a failure in `ESMF_RouteHandleAdvancedUTest`, but I traced that error to I/O, too (in `call ESMF_RouteHandlePrint(rh1, rc=rc)`, which is dying in the call to `IO::destroy(&newIO);` from the `Array::write` function in `ESMCI_Array.C`).

I was also getting a hang in the test `ESMF_FieldSMMFromFileUTest`.

Often, running the tests with clang causes my Mac system to seize up and then do a hard reboot.

(The recently-merged branch `fix/darwin_gfortran_clang` does NOT resolve these issues.)

### LINKED ISSUE #81: NetCDF paths not being picked up properly on Macs with the latest NetCDF from homebrew due to logic in nfconfigtest
With the latest version of NetCDF through homebrew on a Mac, the Fortran includes and libs no longer seem to be bundled with the standard netcdf package. Instead, I needed to install them separately via `brew install netcdf-fortran`. This leads to the Fortran includes and libs being in a different directory from the C includes and libs. nf-config or nc-config can properly tell you about this, but this currently isn't working in an ESMF build with `export ESMF_NETCDF=nc-config` (and so we end up with just the C includes and libs, not the Fortran ones, leading to `Fatal Error: Cannot open module file 'netcdf.mod' for reading`).

The issue seems to be that the script `nfconfigtest` is returning "notworking" even though I have a functional nf-config. That script expects `nf-config --version` to return a string that starts with "netCDF-Fortran", but in the current version of netcdf-fortran through homebrew, `nf-config --version` just gives "4.6.0-development".

@theurich it looks like you added the nfconfigtest script a few years ago, in 8da808d20c. This seems to be trying to detect the possibility that `nf-config` is in your path but somehow broken. Do you remember the details that required you to create this, and do you think we can come up with a different way to detect this situation?

billsacks:
(1) From myself: fixes to get Darwin.gfortranclang combination working. I have tested this with openmpi on my M1 Mac. I will do some other testing today.

(2) From @theurich : Add Linux ARM and LLVM compilers

Resolves #60 
Resolves #81

### DIFF
--- a/scripts/libs.openmpif90_forcxx
+++ b/scripts/libs.openmpif90_forcxx
@@ -1,0 +1,9 @@
+#!/bin/sh
+# This script obtains the Fortran OpenMPI libraries that need to be explicitly added to
+# the link line when linking with a C++ linker. (This adds an extra instance of -lmpi, but
+# that shouldn't be a problem.)
+OPENMPI_MPIF90=`which $1`
+libs=`${OPENMPI_MPIF90} --showme:libs`
+# The above returns the names of the libraries without '-l'. Here we add '-l' to the start
+# of each library name:
+echo $libs | sed -E 's/(^| ) */ -l/g'
--- a/src/addon/NUOPC/src/NUOPC_Driver.F90
+++ b/src/addon/NUOPC/src/NUOPC_Driver.F90
@@ -3448,7 +3448,7 @@ module NUOPC_Driver
   
   !-----------------------------------------------------------------------------
   
-  subroutine routine_executeGridComp(is, i, phase, activeClock, name, userrc, rc)
+  recursive subroutine routine_executeGridComp(is, i, phase, activeClock, name, userrc, rc)
     type(type_InternalState)        :: is
     integer,          intent(in)    :: i, phase
     type(ESMF_Clock), intent(inout) :: activeClock
@@ -3493,7 +3493,7 @@ module NUOPC_Driver
 
   !-----------------------------------------------------------------------------
   
-  subroutine routine_executeCplComp(is, i, j, phase, activeClock, name, userrc, rc)
+  recursive subroutine routine_executeCplComp(is, i, j, phase, activeClock, name, userrc, rc)
     type(type_InternalState)        :: is
     integer,          intent(in)    :: i, j, phase
     type(ESMF_Clock), intent(inout) :: activeClock
--- a/.github/workflows/build-esmf-docs.yml
+++ b/.github/workflows/build-esmf-docs.yml
@@ -51,8 +51,8 @@ jobs:
 
     - name: Copy docs
       run: | 
-        cd ${{ github.workspace }}/esmf-org.github.io
-        mkdir -p docs/nightly/${{ github.ref_name }}
+        cd ${{ github.workspace }}/esmf-org.github.io        
+        mkdir -p docs/nightly/${{ github.ref_name }}/dev_guide
         cd ${{ github.workspace }}/artifacts
         unzip doc-artifacts.zip
         cd ${{ github.workspace }}/artifacts/artifacts/doc-esmf
@@ -62,7 +62,7 @@ jobs:
         cp -rf NUOPC_refdoc.pdf ${{ github.workspace }}/esmf-org.github.io/docs/nightly/${{ github.ref_name }}/
         cp -rf NUOPC_howtodoc ${{ github.workspace }}/esmf-org.github.io/docs/nightly/${{ github.ref_name }}/
         cp -rf NUOPC_howtodoc.pdf ${{ github.workspace }}/esmf-org.github.io/docs/nightly/${{ github.ref_name }}/
-        cd ${{ github.workspace }}/artifacts/artifacts/doc-dev_guide
+        cd ${{ github.workspace }}/artifacts/artifacts/doc-dev_guide        
         cp -rf ./dev_guide/dev_guide/* ${{ github.workspace }}/esmf-org.github.io/docs/nightly/${{ github.ref_name }}/dev_guide/
         
     - name: Commit and publish docs
--- a/.github/workflows/build-esmpy-docs.yml
+++ b/.github/workflows/build-esmpy-docs.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Build Docker image  
       run: | 
         cd ${{ github.workspace }}/esmf-containers/build-esmf-docs/esmpy
-        docker build . --tag esmf/build-esmpy-docs
+        docker build . --tag esmf/build-esmpy-docs --build-arg ESMF_BRANCH="${{ github.ref_name }}" --no-cache
         
     - name: Copy artifacts
       run: |
--- a/.github/workflows/build-esmpy-docs.yml
+++ b/.github/workflows/build-esmpy-docs.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Build Docker image  
       run: | 
         cd ${{ github.workspace }}/esmf-containers/build-esmf-docs/esmpy
-        docker build . --tag esmf/build-esmpy-docs --build-arg ESMF_BRANCH="${{ github.ref_name }}" --no-cache
+        docker build . --tag esmf/build-esmpy-docs
         
     - name: Copy artifacts
       run: |
--- a/build_config/Darwin.gfortran.default/README
+++ b/build_config/Darwin.gfortran.default/README
@@ -1,11 +1,12 @@
 Settings for Darwin (Mac OS X), using the GNU gfortran compiler 
 and GNU C++ compiler (g++).
 
-On some Darwin systems, the clang compiler is provided as an alias to g++.
-This is detected and properly supported.  On systems where both g++ and clang
-exist, ensure that the PATH environment variable is set correctly such that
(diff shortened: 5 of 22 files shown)


scripts/libs.openmpif90_forcxx
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
```
billsacks, line 1: I struggled a bit with what to name this script and am open to other suggestions. I was trying to distinguish this from the existing `libs.openmpif90` (which is still needed for other compiler combinations): both of them accept the mpif90 wrapper executable, which I think is what's indicated by the name `libs.openmpif90`, but the original produces the libraries needed when linking with the f90 compiler, whereas the new one produces the libraries needed when linking with the C++ compiler.

src/addon/NUOPC/src/NUOPC_Driver.F90
```diff
@@ -3448,7 +3448,7 @@ recursive subroutine routine_Run(driver, importState, exportState, clock, rc)
   
   !-----------------------------------------------------------------------------
   
-  subroutine routine_executeGridComp(is, i, phase, activeClock, name, userrc, rc)
+  recursive subroutine routine_executeGridComp(is, i, phase, activeClock, name, userrc, rc)
```
billsacks, line 3451: I needed to add this `recursive` attribute to fix a runtime error when running CTSM. I also added the `recursive` attribute below, for subroutine `routine_executeCplComp`. I'm not sure if that second one was needed. Most subroutines in this module were already declared `recursive`, but please let me know if you think it was wrong for either or both of these to be declared `recursive` (I haven't tried tracing the code to verify that these are really meant to be recursive; my sense is that the recursive calls may arise indirectly via callbacks).

scripts/libs.openmpif90_forcxx
theurich: That name looks fine to me. We haven't been super strict about naming convention of these internally used scripts. The name of this one seems as clear as any other.

src/addon/NUOPC/src/NUOPC_Driver.F90
theurich: No objection. I could easily see how these routines might end up in a recursive call tree. It makes sense for them to carry the `recursive` attribute.

billsacks:
I was delayed in doing some final testing of this due to homebrew updating my netcdf library in a way that broke the ESMF build. I have just tracked this down to #81 so I am hopeful that I can move forward now....

billsacks:
With a workaround for #81 in place, I was able to run the tests on Darwin.gfortranclang with mpich. As with openmpi, there is one failure in ESMF_TraceMPIUTest. In addition, there is a failure with mpich in ESMF_ArraySharedDeSSISTest; I opened #82 to track that issue.

So the overall test results with Darwin.gfortranclang with mpich, with ESMF_TESTEXHAUSTIVE=ON, are:
```
SYSTEM TESTS SUMMARY
Found 49 multi-processor system tests, 48 passed and 1 failed.



EXAMPLES SUMMARY
Found 80 multi-processor examples, 80 passed and 0 failed.



UNIT TESTS SUMMARY
Found 13870 exhaustive multi-processor unit tests, 13869 passed and 1 failed.
```

theurich:
@billsacks this is excellent! The ESMF_TraceMPIUTest is a known issue for other combinations, and will have to be resolved eventually. The ESMF_ArraySharedDeSSISTest failure with OpenMPI is interesting, and I just commented under #82. But for now this will just be another known issue (KI) for the release. 

For the release it would be great if we can set up ESMF regression testing as per https://github.com/esmf-org/esmf-test-scripts/tree/main/python_scripts on your Mac M1 system. It doesn't have to be per cron and automatic, but could be triggered by you manually. Just will be nice to have test reports for the supported platform table. It is basically the basis for us claiming it "supported" in the release.

billsacks:
I finished the one other set of testing I wanted to do on this branch: testing on my old intel-based Mac (with openmpi). Test results look pretty good there, though a number of tests couldn't run due to not having enough processors. In all, with ESMF_TESTEXHAUSTIVE, there were 8 failed system tests and 360 failed unit tests:
- ESMF_TraceMPIUTest fails as before
- The only other real failure is that ESMF_RegridWeightGenUTest dies with a seg fault in PIO
- ESMF_FieldBundleUTest and ESMF_GridCoordUTest were killed by Cortex XDR due to looking like malicious activity
- All other failures were due to not having enough processors on this machine (it has 4 processors)

**As far as I'm concerned, this branch is ready to be merged. @theurich let me know if you'd like me to do anything else before merging.**

theurich (commented):
@billsacks I will turn on automated branch testing. There are a few code changes in this PR, all of them looking good to me, but we need to make sure none of the other combos has issues with them before merging to develop shortly before the freeze. One night should be all e need to see.

theurich (approved):

PR #91, opened 2022-10-18
catch a different pio error message

jedwards4b:
In some cases an attempt to open a netcdf-4 file with pnetcdf will generate a PIO_ENOTBUILT error instead of PIO_EINVAL.   This will capture that error and attempt to reopen with netcdf.

### DIFF
--- a/src/Infrastructure/Mesh/src/ESMCI_Mesh_FileIO.C
+++ b/src/Infrastructure/Mesh/src/ESMCI_Mesh_FileIO.C
@@ -421,7 +421,7 @@ void ESMCI_mesh_create_from_ESMFMesh_file(int pioSystemDesc,
     int mode = 0;
     piorc = PIOc_openfile(pioSystemDesc, &pioFileDesc, &pio_type, filename, mode);
     // if the file was created with netcdf4, it cannot be opened with pnetcdf
-    if (piorc == PIO_EINVAL || PIO_ENOTBUILT){
+    if (piorc == PIO_EINVAL || piorc == PIO_ENOTBUILT){
         pio_type = PIO_IOTYPE_NETCDF;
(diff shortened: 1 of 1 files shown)


theurich:
@jim, is this already part of a more recent PIO release tag? If so, should we just update PIO under ESMF develop?

jedwards4b:
@theurich the change here is in esmf, not in pio.

theurich:
@jim - sorry, I should have looked at the file that changed before making that comment. 
On the PIO topic though, is there a new PIO release tag (with the recent changes) that we should update to under ESMF?

theurich:
@oehmke - please review Jim's PR since it is in Mesh code. Thanks.

jedwards4b:
@theurich no, esmf is on the current pio release.

oehmke (commented):
Line 424 and 428 look like they are dumping error information to stdout. We usually just write error info to the log files. I was wondering if these are left over debug statements. If not, can we take them out and let the PIO error handing on line 430 write the error to the log files instead?

jedwards4b:
@oehmke Sorry, I've removed the debug statements.

oehmke (commented):
Thanks Jim!   Sorry, I just noticed one more thing. I think that you need to compare each variable with a ==. I.e.  (piorc == PIO_EINVAL) || (piorc == PIO_ENOTBUILT)  I think that what you have now does an or of the truth value of (piorc==PIO_EINVAL) with just PIO_ENOTBUILT.

jedwards4b:
@oehmke no worries - good catch.   I am retesting with the change now and will push shortly.

oehmke (approved):
Thanks! Looks good to me now.

theurich:
@jedwards4b and @oehmke - I think this PR got lost in the shuffle. It's merged into develop now and will make it into the next 8.5.0 beta tag.

PR #99, opened 2022-11-08
Potential link options to address Trace failures under Darwin

theurich:
Raffaele provided modified link options on a branch 3 years ago. This was his commit message:
Add Trace link options for ld (LLVM 10.0.0, clang-1000.11.45.5) on Mac OSX (10.13.6).

If this addresses the Trace failures under Darwin, we should consider how this can be integrated into the build system, not adversely affecting other platforms.

### DIFF
--- a/build/common.mk
+++ b/build/common.mk
@@ -1848,9 +1848,9 @@ ESMF_TRACE_WRAPPERS_MPI += mpi_wait_ mpi_wait__ mpi_waitall_ mpi_waitall__
 ESMF_TRACE_WRAPPERS_MPI += mpi_waitany_ mpi_waitany__
 
 COMMA := ,
-ESMF_TRACE_STATICLINKOPTS := -static -Wl,--wrap=c_esmftrace_notify_wrappers -Wl,--wrap=c_esmftrace_isinitialized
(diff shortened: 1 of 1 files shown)


billsacks:
@theurich would you like me to test these changes?

theurich:
I went back in my email, to see if there was any communication about the changes on this branch. I found a thread, and it looks like the changes are probably NOT the right way.

billsacks:
Ah, thanks. So I guess I shouldn't bother testing after all? Or do you feel it's still worth checking?

theurich:
I agree, no action for now!

I will keep this draft open for now, and plan to look into the trace issue again soon as part of the 8.5.0 cycle. It would be nice to resolve the outstanding issues, and the Trace tests are probably the bulk across different platforms at the moment. That's why I was excited when I found this old branch earlier today... and still good to have with the old conversation about it in this PR draft now... and hopefully we can put all this to rest sooner than later.

theurich:
Closing this PR draft. The code on the branch discussed here was already discussed as an improper solution previously. Current work on this issue also points a different direction. To close this now to prevent further confusion.

PR #114, opened 2022-12-15
Unify Regrid RH reuse optimization for NUOPC and ESMF and extend to cover all Grid/Mesh combinations

theurich:
For historical reasons there are currently two RH reuse optimization for Regrid(): one in the NUOPC_Connector and the other in ESMF_FieldBundleRegridStore(). The fundamental reason for this split was that the NUOPC Connector supports different Regrid options for each src/dst-pair, while the ESMF_FieldBundleRegridStore() API does not. Therefore NUOPC_Connector cannot simply call into ESMF_FieldBundleRegridStore(), as it does for Redist() (i.e. ESMF_FieldBundleRedistStore()).

The problem with the current _split_ of the Regrid-RH-Reuse optimization is that any improvement of the reuse optimization would require to update two separate copies of almost the same code, one under NUOPC, and one under ESMF. The two implementations have already started drifting apart e.g. for regrid weight matrix reuse that is implemented under ESMF, but was never carried over to the NUOPC counterpart.

Adding Mesh support in addition to the existing Grid support in Regrid-RH-Reuse  requires that the NUOPC and ESMF implementation must first be unified.

We already worked out a plan on paper to support per-scr/dst-pair options through the ESMF_FieldBundleRegridStore() API, however, implementing this most general approach would take considerable amount of time since it involves public API design and potentially changes, also considering backward compatibility restrictions. Instead the approach chosen under this PR is to unify the implementation via an _internal_ ESMF method called by NUOPC_Connector and ESMF_FieldBundleRegridStore() internally.

The unified routine `ESMF_FieldBundleRegridStorePair()` is located inside `ESMF_FieldBundle.cppF90`. It is Fortran public, but ESMF internal.

Routine `ESMF_FieldBundleRegridStorePair()` implements the Regrid-RH-Reuse optimization general enough to cover both NUOPC and ESMF requirements. It was further extended to also cover all for Grid/Mesh combinations for src/dst-pairs.

### DIFF
--- a/build/common.mk
+++ b/build/common.mk
@@ -1094,7 +1094,7 @@ endif
 ifeq ($(origin ESMF_CLINKRPATHS_ENV), environment)
 ESMF_CLINKRPATHS = $(ESMF_CLINKRPATHS_ENV)
 endif
-ESMF_CLINKRPATHS   += $(ESMF_CXXLINKRPATHS) $(ESMF_F90LINKRPATHS)
+ESMF_CLINKRPATHS   += $(addprefix $(ESMF_CXXRPATHPREFIX),$(subst -L,,$(filter -L%,$(ESMF_CLINKPATHS))))
 # - make sure environment variable gets prepended _once_
 ifeq ($(origin ESMF_CLINKLIBS), environment)
 export ESMF_CLINKLIBS_ENV := $(ESMF_CLINKLIBS)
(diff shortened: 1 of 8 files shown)


theurich:
@feiliuesmf - please set bit-12 for the Connector Verbosity in your runs. As for your original PR, that bit will trigger additional log messages from the optimizer routine. This will let us verify that RHs were only computed when needed. Thanks!

PR #116, opened 2023-01-10
split pytest-json-report into optional `testing` recipe

### LINKED ISSUE #115: ESMPy dependencies on PyPI and Anaconda are inconsistent
Hi ESMF developers,

I maintain a library on PyPI and conda-forge that uses ESMPy and I recently noticed that the conda-forge package for ESMPy appears to be missing a dependency (`pytest-json-report`) that is outlined as needed within the [pyproject.toml](https://github.com/esmf-org/esmf/blob/3d9b652f35039fd1af22ea69606bbbcb4b624fe1/src/addon/esmpy/pyproject.toml#L14). This is an issue as running `pip check` at the testing stage of a build on conda-forge will fail:
```
+ pip check
esmpy 8.4.0 requires pytest-json-report, which is not installed.
```
For the time being, I can add that dependency manually to my library's recipe, but it would make much more sense to list `pytest-json-report` under `optional-dependencies` for testing purposes (see: https://peps.python.org/pep-0633/).

Thanks!

Zeitsperre:
Closes #115

### DIFF
--- a/src/addon/esmpy/pyproject.toml
+++ b/src/addon/esmpy/pyproject.toml
@@ -1,17 +1,16 @@
 [build-system]
-requires = [ "setuptools>=41", "wheel", "setuptools-git-versioning"]
+requires = [ "setuptools>=41", "wheel", "setuptools-git-versioning" ]
 build-backend = "setuptools.build_meta"
 
 [project]
 name = "esmpy"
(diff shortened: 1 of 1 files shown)


rokuingh (approved):
Looks good to me, thanks @Zeitsperre!

PR #121, opened 2023-01-31
Newalarm

danrosen25:
ISSUES:
https://github.com/esmf-org/esmf-support/issues/11
POSSIBLE CONFLICT/DUPLICATE PR:
https://github.com/esmf-org/esmf/pull/29


billsacks:
I have updated this branch to v8.7.0b08, which is the latest point on develop that was easy to update to. (v8.7.0b09 brings in the new repeat capability to the ESMF Clock that @oehmke implemented; this involves conflicts with this branch.)

The list of commits and changed files is much larger than I expect to see in this PR. I'm going to try opening a new PR for this branch in the hopes that it appears cleaner, like I expect.

PR #128, opened 2023-02-18
Fixes #126 - STATEINTENT_INTERNAL

### LINKED ISSUE #126: Request to introduce new constant of type `ESMF_StateIntent_Flag`
MAPL components generally have an "internal" state in addition to import and export states.       Currently ESMF supports:

- `ESMF_STATEINTENT_IMPORT`
- `ESMF_STATEINTENT_EXTERNAL`, and
- `ESMF_STATEINTENT_UNSPECIFIED`

Technically we could exploit `UNSPECIFIED` but that seems a step in the wrong direction.    It would be marginally useful for us to have another named constant for the INTERNAL case.    

We can of course name one ourselves, and that is what is currently done, but then we either (1) perversely export something from MAPL with an ESMF prefix, (2) have a motley collection of constant names, or (3) provide MAPL aliases for the existing ESMF constants:  `MAPL_STATEINTENT_IMPORT` ...

None of these is abhorrent, but (1) would be the best from a "least surprise" point of view.    We would be happy to do the PR depending on what the ESMF team prefers here.

tclune:
This change introduces a new constant `ESMF_STATEINTENT_INTERNAL` of type `ESMF_StateIntent_Flag`.

- Does not attempt to extend/modify built-in tests to cover the (minimal) new functionality.

- Note that the internal ids for the UNSPECIFIED/INVALID cases are changed from 3/4 to 4/5 respectievly.  This could in theory break user code that incorrectly relies on those values.

- Fixed minor typo (missing period) in documentation.

### DIFF
--- a/src/Superstructure/State/src/ESMF_StateAPI.cppF90
+++ b/src/Superstructure/State/src/ESMF_StateAPI.cppF90
@@ -881,7 +881,8 @@ StateAddRepListMacro(ESMF_State,State,nestedStateList,state,nestedStateList(i)%s
 !   \item[{[stateintent]}]
 !    Import or Export {\tt ESMF\_State}.  Valid values are
 !    {\tt ESMF\_STATEINTENT\_IMPORT}, {\tt ESMF\_STATEINTENT\_EXPORT},
-!    or {\tt ESMF\_STATEINTENT\_UNSPECIFIED} The default
+!    {\tt ESMF\_STATEINTENT\_INTERNAL},
+!    or {\tt ESMF\_STATEINTENT\_UNSPECIFIED}. The default
 !    is {\tt ESMF\_STATEINTENT\_UNSPECIFIED}.
 !   \item[{[arrayList]}]
 !    A list (Fortran array) of {\tt ESMF\_Array}s.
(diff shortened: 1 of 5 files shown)


src/Superstructure/State/src/ESMF_StateAPI.cppF90
```diff
@@ -1970,6 +1971,8 @@ type(ESMF_KeywordEnforcer), optional:: keywordEnforcer ! must use keywords below
          msgbuf = "Import State"
        case (ESMF_STATEINTENT_EXPORT%state)
          msgbuf = "Export State"
+       case (ESMF_STATEINTENT_INTERNAL)
```
theurich, line 1974: It looks like this should be `case (ESMF_STATEINTENT_INTERNAL%state)`.

theurich (changes requested):
Besides the comment in code that needs to be addressed, there also should be an update to file `/src/Superstructure/State/doc/State_options.tex` to add the new state intent.

src/Superstructure/State/src/ESMF_StateAPI.cppF90
tclune: Yup.  Thought I could just double check my work and avoid a test compilation in my rush before leaving on travel.   Am back in town and will fix and resubmit tomorrow (I hope).

tclune:
ok - I've verified the latest patch builds on laptop

Issue #94, opened 2022-10-25
Problem with file type detection


oehmke:
The type detection checks for Gridspec first using the type of units. This can cause a scrip file to look like a gridspec file. Probably it would be better to look for specific variables first and then look at units. 

Also, add back in the src and dst file types to ESMF_RegridWeightGen, so a user can force a file type.

billsacks:
The easier piece might be adding back in the src and dst file types as explicit argument... probably start with that.

PR #142, opened 2023-05-13
Re-enable -t, --src_type and --dst_type args in ESMF_RegridWeightGen

billsacks:
Partially addresses esmf-org/esmf#94

I would welcome a review from @oehmke or others.

I have run a bunch of manual tests:
- Ensuring that `-t`, `--src_type` and `--dst_type` all work
- Invalid values for those produce an error
- Specifying `-t` along with `--src_type` or `--dst_type` produces an error
- Explicitly specified type overrides auto-detected type
- Auto-detection of type still works if you don't explicitly specify type
- A warning is printed if explicitly specified type differs from auto-detected type
- The new function that turns a grid file type into a string works correctly for all grid file types allowed by ESMF_RegridWeightGen

I am currently running `gmake all_tests`. Is there any other testing I should do that is specific to ESMF_RegridWeightGen that isn't covered by `gmake all_tests`?

### DIFF
--- a/src/Infrastructure/Util/src/ESMF_UtilTypes.F90
+++ b/src/Infrastructure/Util/src/ESMF_UtilTypes.F90
@@ -1339,6 +1339,7 @@ interface assignment (=)
   module procedure ESMF_ptas2
   module procedure ESMF_ioas
   module procedure ESMF_ifas_string
+  module procedure ESMF_FileFormatAsString
   module procedure ESMF_FileStatusAs
 end interface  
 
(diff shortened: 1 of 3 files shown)


src/Infrastructure/Util/src/ESMF_UtilTypes.F90
```diff
@@ -1930,6 +1931,36 @@ function ESMF_FileFormatNe(FileFormat1, FileFormat2)
...
+     String = 'ESMF_FILEFORMAT_UGRID'
+  else if (FileFormat == ESMF_FILEFORMAT_CFGRID) then
+     ! Note that ESMF_FILEFORMAT_CFGRID is the same as ESMF_FILEFORMAT_GRIDSPEC
+     String = 'ESMF_FILEFORMAT_CFGRID/ESMF_FILEFORMAT_GRIDSPEC'
+  else if (FileFormat == ESMF_FILEFORMAT_MOSAIC) then
+     String = 'ESMF_FILEFORMAT_MOSAIC'
+  else if (FileFormat == ESMF_FILEFORMAT_TILE) then
+     String = 'ESMF_FILEFORMAT_TILE'
+  else
+     String = '(Unexpected ESMF_FILEFORMAT value)'
```
billsacks, line 1961: Note that I don't abort for a bad value because I couldn't see a good way to do this from here, given that the subroutine is used in a `x = y` context (so no opportunity for a return code). Instead I give a particular string in that case. Let me know if there's a better way to handle this.

billsacks:
`gmake all_tests` ran successfully on my Mac (gfortranclang).

PR #143, opened 2023-05-15
Test spack installation through GitHub action

uturuncoglu:
This change introduces a new CI test for Spack ESMF installation:

- It runs tests that are found in [nuopc-app-prototypes](https://github.com/esmf-org/nuopc-app-prototypes) agains the installation to check the functionality.

The successful tests can be seen in here https://github.com/uturuncoglu/esmf/actions/runs/4984695095 which tests both internal and external PIO installation.


uturuncoglu:
@theurich I tested this with GNU and it is working now. I'll sync the branch with esmf develop and test again. Then, I am plaining to work on IntelOne API side to make it work. I now able to ssh to runner which makes me debug the issue easily. I forgot the make this as draft PR and I am not sure how I change it at this point. Anyway, I could close and re open again if you want.

theurich:
@uturuncoglu - great progress! I just changed this PR to a draft.

billsacks:
Thanks for your work on this @uturuncoglu ! I'm going to remove myself as a reviewer because I'm happy with this if you and others are.

uturuncoglu:
@billsacks Okay. That is fine. Still working on Intel One API but maybe that is not too urgent for the next release.

uturuncoglu:
@theurich I think Intel One API implementation is not ready in Spack side. I have couple of issues such as Spack does not pick up correct compiler under GitHub runner and HDF5 dependency is failing by complaining not to find MPI C compiler but same configuration (incl. same Spack configuration files package.yaml and compiler.yaml) is running under my local Docker container. Anyway, I think it is better to look at Intel implementation later. Please let me know what you think. If you think that Intel is not urgent at this point, we could make this PR ready for review. At least, we could regularly test GCC. This PR has also some code related with Intel and if you want we could remove them or keep as it is and fix later.

theurich:
@uturuncoglu I am okay with putting Intel OneAPI under Spack with GitHub action to the side for now. Making this PR ready for review, being GCC focused, sounds like a good thing at this point. Thank you!

uturuncoglu:
@theurich Okay. I have just marked it ready for review.

uturuncoglu:
@theurich I tested new syntax and it seems it is working. The only issue is that following NUOPC app prototypes are failing with 8.4.2 (but pass with 8.5.0b23). Since I am using recent version of NUOPC app prototypes, I think ESMX test are updated. Anyway let me know what you think? I could go with latest snapshot at this point.

```
  FAIL: AtmOcnMedIngestFromConfigProto
  FAIL: AtmOcnPetListProto
  FAIL: ESMX_AtmOcnProto
  FAIL: ESMX_ExternalDriverAPIProto
```

theurich:
@uturuncoglu - that is correct,  those 4 protos were updated to work/test the latest changes in ESMF/NUOPC/ESMX that are available on v8.5.0b23, and would be expected to fail with earlier versions. Clean results for v8.5.0b23 is all we need right now. Great to know the updated Spack syntax is working as expected. Thanks!

I will review this PR as soon as I get to it, so it can be merged into develop.

uturuncoglu:
@theurich Okay. I removed 8.4.2 from the list and testing 8.5.0b23 only (with and without external PIO). Maybe we just need to point develop but it could be sometimes unstable. I think we could update the test once we add new snapshot.

.github/workflows/test-build-spack.yaml
```diff
@@ -0,0 +1,80 @@
+name: Build ESMF Using Spack
+
+on:
+  workflow_dispatch:
```
theurich, line 4: it would be nice to add `inputs` here that allow user to specify ESMF tag to be used when dispatching manually. Default should be `develop`.

.github/workflows/test-build-spack.yaml
```diff
@@ -0,0 +1,80 @@
...
+  push:
```
theurich, line 5: I think `push` should be removed. It is too costly to run this action on every single push. This is more of a nightly test on `develop`, and of course available for manual dispatch.

.github/workflows/test-build-spack.yaml
```diff
@@ -0,0 +1,80 @@
...
+  schedule:
+    - cron: '0 0 * * MON'
+    - cron: '0 0 * * FRI'
```
theurich, line 8: This should be adjusted to run automatically once a day, at around midnight Mountain time on the default `develop` branch. This is part of the ESMF nightly regression testing of `develop`.

.github/workflows/test-build-spack.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+jobs:
+  latest-stable:
```
theurich, line 11: Probably need a different name here, e.g. `spack-test`, since we don't specifically test the latest stable only.

theurich (changes requested):
I added a few change requests via comments. Please take a look. Basically I think we want to make the Spack action part of our nightly `develop` branch testing. At the same time we want to be able to manually dispatch it for any specific tag.

.github/workflows/test-build-spack.yaml
theurich: Also need a field for the nuopc-app-protos tag. Again default to `develop`. But when running this action on a specific ESMF tag/branch, then also the nuopc-app-protros tag/branch needs to be set differently.

.github/workflows/test-build-spack.yaml
```diff
@@ -0,0 +1,80 @@
...
+        ${{ github.workspace }}/.github/workflows/scripts/spack_install.sh \
+          -r ${{ github.workspace }}
+
+    # checkout NUOPC app prototypes
+    - name: Checkout NUOPC app prototypes
+      uses: actions/checkout@v3
+      with:
+        repository: esmf-org/nuopc-app-prototypes
+        path: ${{ github.workspace }}/nuopc-app-prototypes
+        ref: develop
```
theurich, line 72: This is where the nuopc-app-protos tag input would be used.

uturuncoglu:
@theurich do you want to trigger this action from another place? Do I need to convert it to composite action? If this is only for `workflow_dispatch` then I think we could support by just adding inputs to `workflow_dispatch` section. Right?

uturuncoglu:
@theurich I fix couple of issues but I could not run button for running action manually. Do you see any issue with the yaml file?

.github/workflows/test-build-spack.yaml
theurich: I think you need to add `type: string`.

.github/workflows/test-build-spack.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+jobs:
+  spack-test:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        compiler: [gcc@11.3.0]
+        esmf: [${{ github.event.inputs.esmf_version }}+external-parallelio, ${{ github.event.inputs.esmf_version }}~external-parallelio]
```
theurich, line 22: Try changing this to `${{ inputs.esmf_version }}`.

uturuncoglu:
@theurich Okay. I think I fixed it. I am testing now.

uturuncoglu:
@theurich I am not sure what is wrong with this https://github.com/uturuncoglu/esmf/actions/runs/5215668756/jobs/9413651222
since it is failing with `Error: Package 'spack.pkg.builtin.develop' not found.`. There could be an issue in Spack side. I'll keep my on on other runners to see what will happen.

uturuncoglu:
@theurich I think this is because of the syntax. We need to give input as `esmf@develop` not  just `develop` since this information is passed to Spack. Even if I use `esmf@${{ inputs.esmf_version }}` in the matrix definition, we will loose `esmf@=` capability. Does it sound fine to you? As a note, we need to use `develop` etc. for NUOPC app prototypes. Anyway, let me know what you think?

theurich:
I guess we will have to adjust the default accordingly?

uturuncoglu:
@theurich Okay. It is done. The tests are also works without any issue. I have just throw another one after adjusting defaults but I am not expecting any issue.

uturuncoglu:
@theurich do you want to extend description of the version input to include format?

theurich:
> @theurich do you want to extend description of the version input to include format?

Yes, exactly, for both the `esmf_version` and `nuopc_app_version` the `description` should contain hint for what the format is in each case.

theurich:
@uturuncoglu - It looks like the `schedule` triggered action just failed. I think it is the last thing that needs to be addressed before the PR is ready to be merged.

uturuncoglu:
@theurich The cron test is failed. https://github.com/uturuncoglu/esmf/actions/runs/5224301520 I think I know the issue. The inputs that we added to `workflow_dispatch` are not available to cron job since they are used when the action triggered manually. I think we need to find a way to make them available also to cron. Any idea?

theurich:
@uturuncoglu - that looks like _exactly_ what we want to do here!

.github/workflows/test-build-spack.yml
```diff
@@ -0,0 +1,94 @@
...
+        type: string
+        default: 'esmf@develop'
+      nuopc_app_version:
+        description: 'NUOPC Application Prototypes version or tag like develop or v8.5.0b23'
+        required: false
+        type: string
+        default: 'develop'
+  schedule:
+    - cron: '0 6 * * *'
+    - cron: '0 18 * * *'
```
theurich, line 22: @uturuncoglu - We are already past 18:00 UTC. Maybe use `45 18 * * *` to start in a few minutes from now.

.github/workflows/test-build-spack.yml
```diff
@@ -0,0 +1,94 @@
...
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        compiler: [gcc@11.3.0]
+        esmf: ["${{ inputs.esmf_version || env.esmf_version }}+external-parallelio", "${{ inputs.esmf_version || env.esmf_version }}~external-parallelio"]
```
theurich, line 32: I have no idea what the problem is with finding `env`. But it has also surprised me earlier that we have to use `inputs.esmf_version` here, and not `github.event.inputs.esmf_version`. It's like we are already inside the event namespace - and there's no `env` obviously. Maybe try `github.env.esmf_version`?

.github/workflows/test-build-spack.yml
theurich: Sorry, I just reloaded the action status, and see the failed run from earlier... which ran at 18:00 UTC. Having to reload on that page always gets me.

uturuncoglu:
@theurich I am working on creating matrix dynamically since matrix part of the workflow does not support to get variable from env. I found couple of discussion about it and I a trying to implement at this point. I'll let you know when it is ready. For your reference, you could look at following links,

https://brunoscheufler.com/blog/2021-10-09-generating-dynamic-github-actions-workflows-with-the-job-matrix-strategy
https://www.kenmuse.com/blog/dynamic-build-matrices-in-github-actions/

Anyway, if you have any other idea, I could try to implement that one.

uturuncoglu:
@theurich I think I made it. Here are the successful tests:
- manually triggered: https://github.com/uturuncoglu/esmf/actions/runs/5233782382
- cron triggered: https://github.com/uturuncoglu/esmf/actions/runs/5233819981
There was an issue in checking branch of nuopc-app-prototype but I added missed output and it solved the issue.  The successful test (manually triggered) that checkout correct nuopc-app-prototype branch is also in here: https://github.com/uturuncoglu/esmf/actions/runs/5234063661
Sorry about lots of commits since I was testing different approaches. You  might want to squash them together.

theurich:
@uturuncoglu - this is merged into develop now. Nice to have this working! Great work!

PR #144, opened 2023-05-24
Improve auto-detection of file type and related documentation

### LINKED ISSUE #94: Problem with file type detection

billsacks:
(1) Improve auto-detection of file type by ESMF_RegridWeightGen and other tools. Resolves #94 .

(2) Add unit tests of ESMF_IOFileTypeCheck.F90

(3) Improve some documentation of file types


src/Infrastructure/IO/tests/makefile
```diff
@@ -81,3 +84,13 @@ RUN_ESMF_IO_YAMLUTestUNI:
 RUN_ESMF_IO_MultitileUTest:
 	rm -f $(ESMF_TESTDIR)/ESMF_IO_MultitileUTest*.nc
 	$(MAKE) TNAME=IO_Multitile NP=8 ftest
+
+# Note that we run this with NP=1 because there is little value in running these tests
+# with multiple processors
+RUN_ESMF_IO_FileTypeCheckUTest:
+	cp -f filetypecheck_data/*.nc $(ESMF_TESTDIR)
+	$(MAKE) TNAME=IO_FileTypeCheck NP=1 ftest
```
billsacks, line 92: There didn't seem to be value in running this unit test with multiple processors, so I set it up to always run with a single processor. Is this a reasonable thing to do?

billsacks:
Note that the new documentation can be viewed here: https://earthsystemmodeling.org/docs/nightly/filetype_improve_autodetect/ESMF_refdoc/node3.html#SECTION03020000000000000000

oehmke (approved):
I looked into this a bit more. Overall looks great! However, here are a couple of comments: 

  + I now understand that the _TILE format is just the individual tiles of the Gridspec Mosaic format. I skimmed through and see some code for this in Grid create from file, so I’m willing to believe that creating a Grid from just a tile is a case that’s handled in there. I vaguely remember Peggy putting this in a few years ago, so I’m content with that now. 

  + I think that we were trying to move towards _CFGRID (vs. GRIDSPEC) to separate that more from GRIDSPEC mosiac and also to more tightly associate it with just the plain CF format grid.  Would you mind changing the type output from the detector method from  ..._GRIDSPEC to .._CFGRID and also change the table on the top of 12.8 to say CF Grid instead of CF Tile. 

With those, I'm fine for this to go in. (I know that you have a tight turn around with vacation and CESM workshop, so if you don't have time, just merge this and I can try to do it.)

Thanks!

theurich:
@billsacks - it looks like Bob had a few comments and a change request. Do you think this PR could be finished in the next few days? It seems like this one is so close!

billsacks:
Yes, I expect to get to this tomorrow afternoon.

PR #145, opened 2023-05-27
Multi-tile I/O: Support layouts that have something other than 1 DE per PET

### LINKED ISSUE #84: Multi-tile I/O: Support layouts that have something other than 1 DE per PET
One of the remaining limitations of multi-tile I/O is that it only supports 1 DE per PET. We want to relax this constraint.

billsacks:
So far this is a partially-working implementation of allowing other than 1 DE per PET with multi-tile I/O. This works to first order, but some remaining needs are:
- Combining undistributed dimensions with something other than 1 DE per PET leads to failures in the multi-tile case
  - Update (2023-09-20): This is now fixed and covered by unit tests (both for the single-tile and multi-tile cases)
- Need to extend code for handling replicatedDims > 0 with the multi-tile case
  - Update (2023-10-10): Canceling this because I/O with replicatedDims has issues even for the single-tile case
- Need to generalize an error-check
  - Update (2023-10-12): It looks like we can remove this error check (see comment below)

Resolves #84


src/Infrastructure/IO/src/ESMCI_IO.C
```diff
@@ -1197,65 +1197,86 @@ void IO::redist_arraycreate1de(Array *src_array_p, Array **dest_array_p, int pet
...
     distgridToArrayMap = &(distgridToArrayMapVec[0]);
   }
 
-  if ((maxIndexTile[0]-minIndexTile[0]+1)<petCount){
-    ESMC_LogDefault.MsgFoundError(ESMF_RC_INTNRL_BAD,
-      "Index space too small to be distributed across all PETs", ESMC_CONTEXT, rc);
-    return; // bail out
+  if (tileCount == 1) {
+    // FIXME(wjs, 2023-05-26) Generalize this error-check to work with the multi-tile case
+    if ((maxIndexPDimPTile[0]-minIndexPDimPTile[0]+1)<petCount){
```
billsacks, line 1246: I could use some guidance on how to generalize this error check

billsacks:
It turns out that this:

> Combining undistributed dimensions with something other than 1 DE per PET leads to failures in the multi-tile case

is a pre-existing problem even for single-tile fields. I will work on that issue - see #150 .

src/Infrastructure/IO/src/ESMCI_IO.C
billsacks: We talked about this in today's core call. The default distribution distributes along the 0th dimension, which explains the form of the current error check. As for how to extend this to multi-tile: that depends on the default distribution for the multi-tile case. For example, it might be correct to sum the (max - min + 1) for the 0th dimension across all tiles and compare that with the petCount, but that would rely on the default distribution being able to do something like: if tile 1 has 1 row and tile 2 has 10 rows, then we use 1 DE for tile 1 and 10 DEs for tile 2; I'll look into whether that is possible or if - for example - the default distribution requires the same number of DEs for each tile.

billsacks:
It turns out that this:

>  Need to extend code for handling replicatedDims > 0 with the multi-tile case

has issues even for the single-tile case (see #184 ) - so I'm not going to try to address this for multi-tile I/O.

billsacks:
It looks like the error check referred to here:

> * Need to generalize an error-check

can be removed: This error check was added in commit
7379b259f2d2d621e335204026b2e2644318772c (by Gerhard, Jan 12, 2018):
"Add checks to prevent redist_arraycreate1de() from trying to redist an
array where the index space is too small to span all PETs."

Note that there is no error check like this in the case where we're not
doing a redist, which makes me think that this may not be a fundamental
PIO limitation. Furthermore, in a test program, it worked to remove this
error check when writing and reading a 2x4 Field with 10 PETs (and going
through the redist function). So it seems like this check isn't
necessary.

theurich:
When the error check in commit 7379b25 was added in 2018, we were using PIO-1. I am pretty sure at the time PIO had issues with PETs that had no local data defined - and so the check was needed on the ESMF side. We only were checking in case an ESMF Redist() was being utilized, because in that case the distribution was not in the hands of the user. For user distributions that go unchanged into the PIO level, the PIO error message would have been sufficient.

With our change to PIO-2, @billsacks's successful testing without this check, and the technical discussion on the core call, I agree that this check should be removed from the implementation.

billsacks:
@theurich - that makes sense - thank you. With that finalized, I think this branch is finally done, or nearly so... I just want to look back and make sure I have everything I wanted in here.

Issue #136, opened 2023-04-21
Build config fixes for Cygwin


robin-deltares:
At Deltares we use a very old version of ESMF on Windows and I'm trying to build a recent version.

Using the MinGW build config works, but that is i686-based.

Copying some MinGW flags to the Cygwin build config gets me further, but there are some non-functional (ESMF_OS_MinGW) code-switches that still make it fail.

I would like to request permission to create a feature-branch for correcting the Cygwin flags and making the code-switches functional.

theurich:
@robin-deltares best just to make fork, create a branch under your fork, and then issue a PR when ready to be reviewed and merged back into the ESMF develop branch (or maybe even a patch branch for a recent release). Let us know if you have questions.
Thank you for contributing!

robin-deltares:
I've progressed a lot.

Changes:
1. Moved over the Windows-specific build config flags from MinGW to Cygwin.
2. Changed the ESMF_OS_MinGW to OSMF_NO_SIGNALS (simply because they share this setting and I don't know what would be appropriate for each code-block) in the code in a hand full of files.
3. /usr/include/w32api/Winsock.h uses different formats for time, e.g. `error: cannot convert ‘long unsigned int*’ to ‘__ms_u_long*’ {aka ‘unsigned int*’}` Bypassed it with ESMF_NO_SOCKETS for now.

When I've reached a successful build, I'd like to have working code put upstream in a more appropriate manner so that we can build future releases straight from the source, but then I'd need some help.

robin-deltares:
Infrastructure and Superstructure were able to build.

Had some problems with scopes though, e.g.:

src/Infrastructure/Mesh/src/Moab/FileOptions.cpp:58:17: error: ‘strdup’ was not declared in this scope; did you mean ‘strcmp’?

src/Infrastructure/Mesh/src/Moab/GeomQueryTool.cpp:1495:27: error: ‘M_PI’ was not declared in this scope

Will now try to build the whole package.

robin-deltares:
🎉 The suite has been built. Building the tests now.

robin-deltares:
Worked inside of Cygwin, but outside of Cygwin openmpi gives problems with missing /dev/shm and some text-file using a Unix-path.

Retrying build with mpiuni.

robin-deltares:
Success.

```
PS C:\Users\westrene\Documents\esmf_bin> .\ESMF_RegridWeightGen.exe --version
[...]

PR #149, opened 2023-06-13
Fix cygwin build configuration and code switches

theurich:
Start the review process of changes provided by @robin-deltares.

See the associated issue: https://github.com/esmf-org/esmf/issues/136

This PR needs very careful review. A lot of source files have been touched. Best to work on this when having access to a Cygwin system.


DWesl:
You could use GitHub Actions to get a Cygwin installation, I have an example here: https://github.com/DWesl/esmf/tree/esmf-cygwin-compile
It takes two hours to compile and one to run the checks, roughly the same as on a laptop.  It is a bit more difficult to work out which flags guard various definitions (for example, `math.h` defines `M_PI` if you pass `-D_BSD_SOURCE` on the command line).

The doc build isn't working at the moment, and I'm not sure how to fix that.

PR #151, opened 2023-06-23
[ESMpy] Guess esmf.mk locations

### LINKED ISSUE #117: Default mechanism for finding ESMFMKFILE in esmpy
Hi!

This issue is a follow up on conda-forge/esmf-feedstock#91. Since esmpy 8.4.0, importing the module requires the `ESMFMKFILE` environment variable to be set. When using `conda`, this variable gets set upon activation of the conda environment. This is already good, but sadly it doesn't cover all use cases. Some are detailed in the discussion there. In summary, the `activation` step is not always performed in some automated setups or when the python kernel of one environment is accessed by an other one.

I was wondering if in addition to looking for an environment variable, `esmpy` could also search in some "default" locations to avoid failing in the above cases ? I have seen this kind of behaviour in some other packages and it seems to be that on unix platforms it could be easy to set a short list of those locations that could cover most cases. The first `esmf.mk` file found would be used.

For example, in the issue above, I suggest a solution that infers the `.../lib/` folder by looking at where builtin packages are stored. 

What do you think?

I could help in implementing this.

aulemahal:
This fixes #117 and should adress conda-forge/esmf-feedstock#91 and many issues on xESMF.

This adds a "guessing" snippet to `loadESMF.py` for cases when the `ESMFMKFILE` env var is not set.

The new behaviour in 8.4 was to depend on the env variable. The `ESMpy` conda package was updated so that this env var is set upon environment activation. However, there are many cases in which a conda env is used _without_ being activated:
- when used as a kernel in a jupyter lab/notebook instance
- when used to build documentation on ReadTheDocs
- when used as a kernel in some fancy IDEs (PyCharm for example)

This new behaviour generated many issues in xESMF, raised by users who do not really know about the mechanics of ESMF/ESMpy. This PR is in part to help those users, who rely on the "out-of-the-box" aspect of conda.

The two paths I added to the guessing are the same that `ESMpy` already guesses in its conda recipe. See: https://github.com/conda-forge/esmpy-feedstock/blob/60527fe641e89da9e2ab0b3f93a5b6f67a0d20f8/recipe/scripts/activate.sh#L11-L15

Thus I am suggesting replicating the "guessing" part from ESMpy's conda configuration directly in its code.

I also removed a few lines from the documentation. Unless I am mistaken, installing ESMpy from source does _not_ require the `esmf.mk` file anymore, no?

### DIFF
--- a/src/addon/esmpy/doc/install.rst
+++ b/src/addon/esmpy/doc/install.rst
@@ -62,15 +62,7 @@ Installing ESMPy from Source
 ----------------------------
 
 When installing from source, ESMPy uses `pip <https://pypi.org/project/pip//>`_ 
-to build and install the package. This requires setting an environment variable 
-pointing to a file named esmf.mk that is generated during an ESMF installation.  
-The path of this file is:
-
-.. code::
-
-    <ESMF_INSTALL_DIR>/lib/lib<g<or>O>/<platform>/esmf.mk
-
-An installation of ESMPy in the default location for Python packages can be done
+to build and install the package. An installation of ESMPy in the default location for Python packages can be done
 with the following command issued from the top level ESMPy directory (``src/addon/esmpy``):
 
 .. code::
@@ -93,7 +85,8 @@ To use ESMPy in an external program, import it with:
 
     import esmpy
 
-The environment variable ``ESMFMKFILE`` must be set when using ESMPy.
+The environment variable ``ESMFMKFILE`` should be set when using ESMPy. If it is not found, the package will
+try to guess a few very common locations, but we recommend correctly setting the variable nonetheless.
 
 .. Note::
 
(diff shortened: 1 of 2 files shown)


rokuingh (approved):
Looks great, thanks for this excellent contribution @aulemahal!

theurich:
@rokuingh and @billsacks - should this still go into release/8.5.0? If you think so, it is okay with me. Just change the target of this PR and let me know. Thanks.

billsacks:
I would vote for including this in the 8.5.0 release since it's tripping up users of ESMPy. @rokuingh do you agree?

kadircs:
I installed from source and I got this error:
```
ImportError: The esmf.mk file cannot be found. Pass its path in the ESMFMKFILE environment variable.
```

billsacks:
@kadircs - despite the guesses that were helpfully added in this PR, in many / most situations when installing from source, you still need to explicitly set the ESMFMKFILE environment variable (see https://earthsystemmodeling.org/esmpy_doc/release/latest/html/install.html#importing-esmpy). If you don't know where `esmf.mk` is, you can run `find . -name esmf.mk` from the ESMF install location.

PR #106, opened 2022-11-29
Modify ESMX build utilities

danrosen25:
These changes enhance the user experience when using ESMX by integrating the component build step into the ESMX build. In the best case scenario the component projects builds and installs libraries using cmake and exports a CMake configuration file including the NUOPC cap. Several other options are available for searching for libraries and fortran modules. Note that external libraries without a CMake configuration file can be particularly challenging and make use of the link\_libraries and link\_paths options. link\_paths accepts both a full path or an environment variable.

### Build Types
**Auto**<br>
The ESMX build system searches for CMakeLists.txt, Makefile, and a build\_script in order in the source\_dir and uses the first build option it finds. If no build files are found then the ESMX build system searches for the CMake configuration file, fortran module, and libraries in the install\_prefix directory but does not build the model. If no build option, CMake configuration, or libraries are found then the build fails.

**CMake**<br>
The ESMX build system searches for CMakeLists.txt in the source\_dir and builds using CMake. Once built, the ESMX build system searches for the CMake configuration file and libraries in the install\_prefix directory.

**Make**<br>
The ESMX build system searches for Makefile in the source\_dir and builds using Make. Once built, the ESMX build system searches for libraries and fortran modules in the install\_prefix directory.

**Script**<br>
The ESMX build system searches for a build\_script in the source\_dir and builds using this script. Once built, the ESMX build system searches for libraries and fortran modules in the install\_prefix directory.

**None**<br>
The ESMX build system will not build the component. The ESMX build system searches for the CMake configuration file, fortran module, and libraries in the install\_prefix directory.

See documentation here [https://github.com/esmf-org/esmf/tree/feature/esmx_cmake/src/addon/ESMX](https://github.com/esmf-org/esmf/tree/feature/esmx_cmake/src/addon/ESMX)

see examples here: [https://github.com/esmf-org/nuopc-app-prototypes/tree/feature/esmx_cmake](https://github.com/esmf-org/nuopc-app-prototypes/tree/feature/esmx_cmake)

Changes
* adds ESMX_Builder script
* add ESMX_BUILD_FILE argument (default esmxBuild.yaml)
* removes configuration option requirements
* add build configuration options (see table below)

| Option          | Description                                       | Default                |
| --------------- | ------------------------------------------------- | ---------------------- |
| build\_type     | auto, cmake, make, script, none                   | auto                   |
| build\_script   | build script                                      | compile                |
| build\_args     | build arguments passed to ExternalProject         | *None*                 |
[...]

PR #152, opened 2023-06-27
Improve HConfig/ESMX integration and associated clean-up and improvements

theurich:
This PR is aimed at branch feature/esmx_cmake. It brings in a few changes that should be part of the final PR #106.

Overview of changes:
- Implementation of `ESMF_HConfigValidateMapKeys()`, and usage from ESMX.
- `ESMX_Data` minor changes to robustify.
- Minor doc changes.

### DIFF
--- a/src/Infrastructure/FieldBundle/src/ESMF_FieldBundle.cppF90
+++ b/src/Infrastructure/FieldBundle/src/ESMF_FieldBundle.cppF90
@@ -4720,6 +4720,9 @@ call ESMF_LogWrite("found rhListMatch -> reuse routehandle!", &
     logical                         :: rhListMatch
     type(ESMF_RouteHandle)          :: rhh
     logical                         :: verbosityFlagIntl
+    character(ESMF_MAXSTR)          :: fName
+    integer                         :: dimCount, rank
+    character(800)                  :: msgString
 
     verbosityFlagIntl = .false. ! default
     if (present(verbosityFlag)) verbosityFlagIntl = verbosityFlag
@@ -4733,6 +4736,27 @@ call ESMF_LogWrite("found rhListMatch -> reuse routehandle!", &
     if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
       ESMF_CONTEXT, rcToReturn=rc)) return
 
+    if (verbosityFlagIntl) then
+      call ESMF_FieldGet(srcField, name=fName, dimCount=dimCount, rank=rank, &
+        rc=localrc)
+      if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
+        ESMF_CONTEXT, rcToReturn=rc)) return
+      write(msgString,'(4A,I2,A,I2)') ESMF_METHOD, ": srcField: ", trim(fName), &
+        " dimCount=", dimCount, " rank=", rank
+      call ESMF_LogWrite(msgString, ESMF_LOGMSG_DEBUG, rc=localrc)
+      if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
+        ESMF_CONTEXT, rcToReturn=rc)) return
+      call ESMF_FieldGet(dstField, name=fName, dimCount=dimCount, rank=rank, &
+        rc=localrc)
+      if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
+        ESMF_CONTEXT, rcToReturn=rc)) return
+      write(msgString,'(4A,I2,A,I2)') ESMF_METHOD, ": dstField: ", trim(fName), &
+        " dimCount=", dimCount, " rank=", rank
+      call ESMF_LogWrite(msgString, ESMF_LOGMSG_DEBUG, rc=localrc)
+      if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
+        ESMF_CONTEXT, rcToReturn=rc)) return
+    endif
+
     rhReuse = ((srcGeomtype==ESMF_GEOMTYPE_GRID) .or. &
                (srcGeomtype==ESMF_GEOMTYPE_MESH))
     rhReuse = rhReuse .and. ((dstGeomtype==ESMF_GEOMTYPE_GRID) .or. &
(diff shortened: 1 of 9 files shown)


theurich:
@danrosen25 - once you approve these changes into branch `feature/esmx_cmake`, I think we are ready to finish up PR #106, into `develop` right after. This is the last I am planning before the freeze.

src/Infrastructure/HConfig/interface/ESMF_HConfig.F90
```diff
@@ -13589,6 +13589,102 @@ subroutine ESMF_HConfigSetMapValStringSeq(hconfig, content, keywordEnforcer, &
...
+      ESMF_CONTEXT, rcToReturn=rc)) return
+
+    ! iterate over hconfig validating each key
+    ESMF_HConfigValidateMapKeys = .true.
+    hconfigIter = hconfigIterBegin
+    do while (ESMF_HConfigIterLoop(hconfigIter, hconfigIterBegin, &
+      hconfigIterEnd, rc=localrc))
+      if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
+        ESMF_CONTEXT, rcToReturn=rc)) return
+      key = ESMF_HConfigAsStringMapKey(hconfigIter, rc=localrc)
```
danrosen25, line 13670: Is this going to work with all compilers? I haven't used the automatic allocation of dynamic length strings.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
```diff
@@ -148,6 +153,45 @@ subroutine SetServices(xdata, rc)
...
+        line=__LINE__,    file=__FILE__)) return  ! bail out
+      call ESMF_ConfigGet(config, hconfig=hconfig, rc=rc)
+      if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, &
+        line=__LINE__, file=__FILE__)) return  ! bail out
+      hconfigNode = ESMF_HConfigCreateAt(hconfig, keyString=compLabel, rc=rc)
+      if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, &
+        line=__LINE__, file=__FILE__)) return  ! bail out
+      ! component responsibility to validate ESMX handled options here, and
+      ! potentially locally handled options
+      isFlag = ESMF_HConfigValidateMapKeys(hconfigNode, &
```
danrosen25, line 174: I'm okay with this but it will have to evolve as the component configuration keys for the driver evolves, that could be tricky. It also ties ESMX Data to ESMX Driver more than I had wanted.

src/addon/ESMX/README.md
```diff
@@ -165,7 +167,7 @@ This section contains a key for for each *component-name*, specifying component
 | `test_exe`       | executable used to run test case              | ESMX_TEST_EXE          |
 | `test_tasks`     | number of tasks used to run test case         | ESMX_TEST_TASKS        |
 
-**Caution:** Downloading components using the `git_repository` option will result in a detached head. Developers making changing to component code must create or checkout a branch before making code changes. Downloading component using git_repository fails if the source_dir already exists.
+**CAUTION:** Downloading components using the `git_repository` option will result in a detached head. Developers making changing to component code must create or checkout a branch before making code changes. Downloading component using git_repository fails if the source_dir already exists.
```
danrosen25, line 170: Oh, I think I fixed this actually by using explicit calls to git within the ESMX Driver CMakeLists.txt code.

danrosen25 (commented):
I can approve this as is with a little more understanding of dynamic length Fortran strings. I have looked at using them before but I can't find clear information about allocation and I was wondering if it varies by compiler or compiler version. Also there are some cosmetic changes.

- do you to allocate memory for the string in user code?
- do you need to deallocate memory in user code?
- if you're not handling allocating in user code what happens when you reuse the same string?

src/Infrastructure/FieldBundle/src/ESMF_FieldBundle.cppF90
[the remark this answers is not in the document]
theurich: I is from me working on using `ESMX_Data` under the `ESMX_AtmOcnProto`. I had some issues at first, where the RegridStore() would bail out under the Connector. To understand this better, I added additional logging output when Verbosity is set... so it was work that happened here and is good to have.

src/Infrastructure/HConfig/examples/ESMF_HConfigEx.F90
```diff
@@ -430,6 +430,7 @@ program ESMF_HConfigEx
 ! the {\em map key} or {\em map value}, respectively.
 !EOE
 !BOC
+  ! type(ESMF_HConfigIter) :: hconfigIter
```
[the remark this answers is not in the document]
theurich: Yes, that is how it should be. It is a comment that shows up as code comment in the example/doc, indicating to the user what type `hconfigIter` is.

src/Infrastructure/HConfig/interface/ESMF_HConfig.F90
```diff
@@ -13589,6 +13589,102 @@ subroutine ESMF_HConfigSetMapValStringSeq(hconfig, content, keywordEnforcer, &
...
+  function ESMF_HConfigValidateMapKeys(hconfig, vocabulary, keywordEnforcer, &
+    badKey, rc)
+!
+! !RETURN VALUE:
+    logical :: ESMF_HConfigValidateMapKeys
+!
+! !ARGUMENTS:
+    type(ESMF_HConfig),        intent(in)            :: hconfig
+    character(len=*),          intent(in)            :: vocabulary(:)
+type(ESMF_KeywordEnforcer), optional:: keywordEnforcer ! must use keywords below
```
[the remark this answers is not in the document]
theurich: Are you wondering if line 13608 should have some whitespace to line up? It does not, because the ProTex processor picks it up without whitespace, and inserts the actual language for optional args requiring keywords.

src/Infrastructure/HConfig/interface/ESMF_HConfig.F90
theurich: Yes, the HConfig API is written to use dynamic character variables. It works on all the compilers we support.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
theurich: Yes, I was thinking about this, too. In the end I did not see any other way than to say, it is the components responsibility to check the validity of all of the keys specified under *its* section. Only the component itself knows the full list of valid keys. Of course a component can be sloppy and not validate the keys... it will work, but might not catch user errors like typos.

src/addon/ESMX/README.md
theurich: I'd say, lets merge this in still as is, but then you remove that paragraph under `feature/esmx_cmake` afterward.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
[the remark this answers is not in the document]
theurich: Yes, that is possible. Again, let's merge this, and then you modify under `feature/esmx_cmake`.

theurich:
@danrosen25 - the `character(:), allocatable`, it works on all of the compilers that we support. The HConfig API uses it for strings. Fortran `character(:), allocatable` automatically allocate on assignment. You can also explicitly allocate, but automatic allocation on assignment is the convenient way. It automatically deallocates when going out of scope. It also automatically re-allocates when re-assigned.

PR #153, opened 2023-06-27
Fix ESMF_FieldGet localElementCount

bena-nasa:
This allows one to get the local element count for multi-tile grids rather than just failing as there seems to be no reason the LOCAL arguments to field get can not work on multi tile grids. I broke up the if present block for the get global and get local arguments into 2. The get global still fail if the number of tiles is > 1, but the localElementCount etc now work for multi-tile grids (aka the cubed-sphere for example).

### DIFF
--- a/src/Infrastructure/Field/src/ESMF_FieldGet.cppF90
+++ b/src/Infrastructure/Field/src/ESMF_FieldGet.cppF90
@@ -631,9 +631,7 @@ type(ESMF_KeywordEnforcer), optional:: keywordEnforcer ! must use keywords below
     endif
 
     if(present(minIndex) .or. present(maxIndex) .or. &
-      present(elementCount) .or. &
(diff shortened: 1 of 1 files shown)


billsacks:
Thanks a lot for submitting this change! Your changes look good to me, but in reviewing the code, I saw some potential for code cleanup (removing variables not needed in certain blocks of the conditional, and consolidating duplicated blocks of code... some of this is pre-existing stuff from before your PR, but I figure this is a good time to clean it all up, since the split into two separate blocks makes some of this more apparent). I'm working on this and will push my changes once they're done – probably next week.

@bena-nasa - Is it okay with you if I push directly to your branch, or would you prefer that I open a separate PR to keep your branch unchanged?

billsacks:
I have finished doing the refactoring I wanted to do - on a branch off of this branch. I am running full testing now and will merge to develop once that finishes cleanly.

Issue #134, opened 2023-04-04
ESMF, CMake and Spack: Need for possible change with Open MPI (and other MPI??)


mathomp4:
Apologies to @theurich, this is going to be a long one.

So, one of my goals is to get GEOS/MAPL to build with Spack (mainline), and I am close. The current issue is with ESMF. So, I can do all my `spack install` commands and all is good and I can `spack load` or `module load` the resulting installations.

However, when I try and build MAPL with:
```
cmake .. -DCMAKE_BUILD_TYPE=Debug -DUSE_F2PY=OFF -DCMAKE_INSTALL_PREFIX=../install-Debug -DBUILD_WITH_FLAP=NO
```
it fails at link time with (for example):
```
❯ make MAPL_cfio_r4
[  0%] Building Fortran object MAPL_cfio_r4/CMakeFiles/MAPL_cfio_r4.dir/ESMF_CFIOBaseMod.f.o
[  0%] Building Fortran object MAPL_cfio_r4/CMakeFiles/MAPL_cfio_r4.dir/ESMF_CFIOUtilMod.F90.o
[  0%] Building Fortran object MAPL_cfio_r4/CMakeFiles/MAPL_cfio_r4.dir/ESMF_CFIOGridMod.F90.o
[  0%] Building Fortran object MAPL_cfio_r4/CMakeFiles/MAPL_cfio_r4.dir/ESMF_CFIOVarInfoMod.F90.o
[  0%] Building Fortran object MAPL_cfio_r4/CMakeFiles/MAPL_cfio_r4.dir/ESMF_CFIOFileMod.F90.o
[  0%] Building Fortran object MAPL_cfio_r4/CMakeFiles/MAPL_cfio_r4.dir/ESMF_CFIOSdfMod.F90.o
[ 50%] Building Fortran object MAPL_cfio_r4/CMakeFiles/MAPL_cfio_r4.dir/ESMF_CFIOMod.F90.o
[ 50%] Building C object MAPL_cfio_r4/CMakeFiles/MAPL_cfio_r4.dir/ShaveMantissa.c.o
[100%] Linking Fortran shared library ../lib/libMAPL_cfio_r4.dylib
ld: library not found for -lmpi_usempif08
collect2: error: ld returned 1 exit status
make[3]: *** [lib/libMAPL_cfio_r4.dylib] Error 1
make[2]: *** [MAPL_cfio_r4/CMakeFiles/MAPL_cfio_r4.dir/all] Error 2
[...]

PR #159, opened 2023-07-07
Also add mpif90 libdirs to ESMF_F90LINKLIBS for Darwin.gfortranclang

billsacks:
This is needed for the unusual case where an application is being linked without the mpi compiler wrappers.

Resolves esmf-org/esmf#134 -- see that issue for details

I'm running testing on my Mac now to confirm that this doesn't break anything.
- Update: testing passes other than the one expected failure on my Mac

@theurich please let me know if you are okay with this change. @mathomp4 please let me know if this fixes the issue for you.


billsacks:
Testing all passes on my Mac with this change (other than the one expected failure).

theurich (approved):
@billsacks - looks okay to me. Thanks for fixing and testing!

Please merge into develop.

PR #161, opened 2023-07-13
Fix ESMX git update for existing component

danrosen25:
Fixes bug in calling git for ESMX component that already has source directory.


theurich (approved):
Looks clean.

PR #162, opened 2023-07-13
Fix scope for undefined ESMX_EXE_NAME

danrosen25:
Fix the scope for ESMX_EXE_NAME when exe_name is not set in esmxBuild.yaml file.


theurich (approved):
I checked with the CMake documentation, and indeed, it does need both calls to set the variable in the local _and_ parent scope. Looks good now.

PR #163, opened 2023-07-13
ESMX/CMake top level to set the compilers as per esmf.mk

theurich:
This addresses the issue observed e.g. under https://github.com/esmf-org/esmf-test-artifacts/blob/d4c27618861e8f7e8041036e6bda5361772b7151/release_8.5.0/intel/18.0.5/O/intelmpi/2018.4.274/out/nuopc.log where the ESMX Proto cases do not build correctly.

To address this issue, correctly set `CMAKE_*_COMPILER` variables on the top level `ESMX/CMakeLists.txt` file. 

Also update minimum ESMF version required to 8.5.0 to be consistent with the latest ESMX version.

Finally also set ESMX version to 1.0.0 in the project() command.


danrosen25:
This looks good. Note that if the developer is integrating the ESMX Driver into their component they'll skip over the application layer and they will be responsible to set CMAKE_Fortran_COMPILER to match their ESMF build. I think this is acceptable. I'm not going to test these changes because I don't have a reproducer but they look good.

src/addon/ESMX/Comps/ESMX_Data/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.19)
```
danrosen25, line 1: Do you want to update this to 3.22 for consistency?

danrosen25 (changes requested):
Update all cmake_minimum_required in ESMX to require version 3.22 in all CMakeLists.txt files. And update the README file to say that 3.22 is required.

theurich:
Yes, correct, for user level project integration, the user `CMakeLists.txt` file is responsible. I made that change in the `ESMX_ExternalDriverAPIProto` and tested locally and on Hera. I will update the protos at the same time when this PR comes in.

I am working on updating the section in the `ESMX/README.md` right now that talks about project integration. Will push to the same branch in just a second.

src/addon/ESMX/Comps/ESMX_Data/CMakeLists.txt
theurich: I was thinking of keeping local minimum version requirements as where they really are... but thinking about it, because of the integration into ESMX that depends on 3.22 now, we might never know when 3.19 might really not be correct. So, I think it is best to bump them all to 3.22 now. Thanks, I will push that with the README.md update when I have it.

theurich:
@danrosen25 - all my follow-on changes are in. Please take another a look.

danrosen25 (approved):

theurich:
@danrosen25 - I just noticed a typo in the README.md change... typing too fast. Small thing so I won't have you review again. Moving forward with the merge. Thanks!

PR #164, opened 2023-07-17
ESMF doc updates + user doc section on Intel Compiler Support

theurich:
This PR covers a few doc updates:

- Formatting of front page of all the ESMF docs to prevent vertical overflow.
- Consistent use of `make` vs `gmake` in docs.
- New [User's Guide section "Intel Compiler Support"](https://earthsystemmodeling.org/docs/nightly/feature/work-on-docs/ESMF_usrdoc/node10.html#SECTION000106100000000000000), specifically covering Classic and oneAPI issue.


src/doc/ESMF_quickstart.tex
```diff
@@ -174,19 +174,19 @@ \subsubsection{Environment variables}
 
 \subsubsection{GNU make}
 The ESMF build system uses the GNU make program; it is normally named 
-{\tt gmake} but may also be simply {\tt make} or {\tt gnumake} on some 
-platforms (we will use {\tt gmake} in this document). ESMF does not use
+{\tt make} but may also be {\tt gmake} or {\tt gnumake} on some 
+platforms. We will use {\tt make} in this document. ESMF does not use
```
billsacks, line 178: I see similar wording in user_make.tex, which gets included in the "Building the ESMF Library" section of the "Building and Installing the ESMF" chapter. I'm thinking that should be changed, too, for consistency (e.g., adding, "We will use {\tt make} in this document." In fact, would it make sense to remove the first few sentences here and replace them with an inclusion of user_make.tex so that this piece stays consistent between the installation instructions and the quick start?

billsacks, line 178: After "We will use {\tt make} in this document", I'm wondering if it would be worth explicitly adding something like, "but note that you may need to replace this usage with {\tt gmake} or {\tt gnumake} to ensure the use of the appropriate make program on your system". I know this kind of repeats what is already said, but it feels like this is something that could trip people up (it has definitely tripped me up in other contexts in the past), so I wonder if it's worth being really explicit here.

billsacks (changes requested):
The new section is well-written (as always!). I appreciate the care put into being consistent between the usage of `make` vs. `gmake`. I have a couple of suggestions for ways to improve this further, but I don't feel strongly about either of my suggestions if you want to leave this as is.

src/doc/ESMF_buildinstall.tex
danrosen25, line 1: I can't tell where in the documentation this section exists. I just don't see it.

danrosen25 (approved):
There's a lot of changes from gmake to make but I looked at the pdf artifact and the only gmake text I see left says something similar to 'use make but on some systems it's gmake or gnumake'

README.md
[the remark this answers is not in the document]
theurich: Yes, and I think it is okay. One at the top, the other at the bottom of the `README.md`.

src/doc/ESMF_buildinstall.tex
theurich: I am looking into this right now. It looks like an unused file. I am testing locally with it gone to see if anything in the doc breaks, and if not will suggest removing the file altogether. Thanks!

src/doc/ESMF_quickstart.tex
theurich: I just reworked this, using `user_make.tex` in both places that discuss the GNU Make dependency. Thanks for pointing this out.

theurich: I agree that this could be more explicit. I just pushed a change to better clarify this in `user_make.tex`, which now is used where the GNU Make dependency is discussed. Please review once the online branch docs have rebuilt. Thanks!

theurich:
All, the online docs are up-to-date with my most recent pushes. @danrosen25 specifically, please see the following sections to address the issues you raised. Thanks!
- https://earthsystemmodeling.org/docs/nightly/feature/work-on-docs/ESMF_usrdoc/node6.html#SECTION00063200000000000000
- https://earthsystemmodeling.org/docs/nightly/feature/work-on-docs/ESMF_usrdoc/node10.html#SECTION000107000000000000000
- https://earthsystemmodeling.org/docs/nightly/feature/work-on-docs/ESMF_usrdoc/node10.html#SECTION000105000000000000000
- https://earthsystemmodeling.org/docs/nightly/feature/work-on-docs/ESMF_usrdoc/node10.html#SECTION000106000000000000000

billsacks (approved):
This looks great - thanks @theurich ! I'm happy with the wording for the `make` changes, and  I also like the changes you made for describing which environment variables are (not necessarily) needed (e.g., ESMF_COMPILER on Darwin systems).

src/doc/ESMF_install.tex
```diff
@@ -1111,6 +1111,37 @@ \subsection{Supported Platforms}
 will be stored in a separate directories so the results will be kept
 separate for different architectures or options.
 
+\subsubsection{Intel Compiler Support (Classic and oneAPI)}
+
+ESMF supports the Intel compiler suite via {\tt ESMF\_COMPILER=intel}. Starting in 2020, Intel began promoting their
+new LLVM-based C/C++ and Fortran compiler line under the oneAPI brand. As of 2023, the older compilers are still
+part of the Intel compiler suite and referred to as Intel Compiler Classic. ESMF supports both the classic and oneAPI
+compiler lines. The following paragraphs povide important details that allow users to fine-tune the interaction
```
oehmke, line 1119: "povide" should be "provide"

oehmke (commented):
Aside from the one misspelling I marked in ESMF_install.tex. It looks good. It's nice to have things switched gmake to make.

src/doc/ESMF_install.tex
theurich: Fixed. Thanks!

Issue #150, opened 2023-06-16
Problems reading a field with ungridded dimensions and > 1 DE per PET


billsacks:
In working on multi-tile I/O, I discovered a pre-existing problem for the standard, single-tile I/O: When trying to read a field that has both ungridded dimensions and > 1 DE per PET, the read fails like this, on PETs that have > 1 DE per PET:

```
20230616 133812.347 ERROR            PET2 ESMCI_Array.C:756 ESMCI::Array::create() Invalid argument  - Mismatch in localDeCount between larrayList argument and DELayout
20230616 133812.347 ERROR            PET2 ESMCI_IO.C:1407 ESMCI::IO::undist_arraycreate_al Invalid argument  - Internal subroutine call returned Error
20230616 133812.347 ERROR            PET2 ESMCI_IO.C:349 ESMCI::IO::read() Invalid argument  - Internal subroutine call returned Error
```

The write works for a field like this, and the read works for fields with either ungridded dimensions or > 1 DE per PET but not both. My guess is that the order of dealing with ungridded dimensions and > 1 DE per PET is wrong for reads. I will investigate this.

PR #169, opened 2023-07-25
On read, change order of dealing with redist vs undist

billsacks:
The code has not properly handled the read of a Field or Array with both ungridded / undistributed dimensions and > 1 DE per PET. It appears that this is due to an incorrect ordering of dealing with these two situations in IO::read. This commit changes that ordering to resolve this issue.

Note that I also needed to reassign temp_array_p to temp_array_undist_p before doing the redistStore and redist; without this, the redistStore call generated an error with this trace-back:

```
20230724 170325.547 ERROR            PET3 ESMCI_Array.C:6238 ESMCI::Array::tRedistStore() Invalid argument  - srcArray and dstArray must provide identical number of exclusive elements
20230724 170325.547 ERROR            PET3 ESMCI_Array.C:6018 ESMCI::Array::redistStore() Invalid argument  - Internal subroutine call returned Error
20230724 170325.547 ERROR            PET3 ESMCI_IO.C:387 ESMCI::IO::read() Invalid argument  - Internal subroutine call returned Error
20230724 170325.547 ERROR            PET3 ESMCI_IO.C:282 ESMCI::IO::read() Invalid argument  - Internal subroutine call returned Error
20230724 170325.548 ERROR            PET3 ESMCI_IO_F.C:210 c_esmc_ioread() Unable to read from file  - Internal subroutine call returned Error
20230724 170325.548 ERROR            PET3 ESMF_IO.F90:397 ESMF_IOAddArray() Unable to read from file  - Internal subroutine call returned Error
20230724 170325.548 ERROR            PET3 ESMF_FieldPr.F90:336 ESMF_FieldRead Unable to read from file  - Internal subroutine call returned Error
20230724 170325.548 ERROR            PET3 esmApp.F90:185 Unable to read from file  - Passing error in return code
```

Resolves esmf-org/esmf#150

I have verified that a round-trip write-then-read of a single-tile Field with both ungridded dimensions and 0 or > 1 DE per PET on some PETs gives identical results to the original Field.

@theurich and/or @oehmke - I would like your input on two things here (I am assigning you both as reviewers, but input from one of you is probably sufficient):

(1) I would like a second set of eyes / brain on the logic here to confirm that this flow of logic looks reasonable. Specifically, I want confirmation that the logic around the redistStore call looks reasonable. At a high level, for an Array with both undistributed dimensions and > 1 DE per PET, we (a) create a temp Array with 1 DE per PET; (b) create an aliased version of this Array that treats all dimensions as distributed, saving the result of (a) as `temp_array_undist_p`; (c) Read into the Array created in (b); (d) do a redistStore and then redist from `temp_array_undist_p` into the original (user) Array. I *think* we can rely on `temp_array_undist_p` being filled properly with data by the Read into `temp_array_p` because of the reference sharing between `temp_array_undist_p` and `temp_array_p` (and my testing gives me more confidence in this), but I want to check to make sure you don't see any edge cases where this logic will fail. (I had originally hoped that I could do the redistStore and redist directly from `temp_array_p` back to the original (user) Array, but I wasn't too surprised to find that this failed, with the message given in the above traceback.)

(2) I did this testing with a standalone program (https://github.com/billsacks/esmfprojects-multi_tile_io/blob/main/esmApp.F90), not a unit test. I plan to add unit testing of this complex case (Array / Field with both ungridded dimensions and 0 or > 1 DE per PET on some PETs) in my multi-tile IO unit tests. My question is: do you feel I should also add unit tests of this combination on single-tile Arrays / Fields – probably in ESMF_ArrayIOUTest and ESMF_FieldIOUTest? It seems like we'll get the necessary code / logic coverage through the multi-tile tests, but the advantage of testing this combination with single-tile Arrays / Fields is that, if the test fails in the future, it would allow distinguishing between an issue with multi-tile vs. an issue that also impacts single-tile. The downside is that I think setting up these tests will be a bit complex, so adding single-tile versions of these tests will mean more test code to write and maintain. I can go either way on this myself; do you have feelings?

### DIFF
--- a/src/Infrastructure/Field/tests/ESMF_FieldIOUTest.F90
+++ b/src/Infrastructure/Field/tests/ESMF_FieldIOUTest.F90
@@ -65,6 +65,7 @@ program ESMF_FieldIOUTest
   type(ESMF_Mesh) :: elem_mesh
   type(ESMF_Field) :: field_att, field_ugd_att
   type(ESMF_Field) :: field_ug, field_ug2
+  type(ESMF_Field) :: field_ug_w2DE, field_ug_r2DE
 
   type(ESMF_DistGrid) :: dg_debl
   type(ESMF_DistGridConnection), allocatable :: connectionList(:)
@@ -75,6 +76,8 @@ program ESMF_FieldIOUTest
 
   real(ESMF_KIND_R8), pointer :: Farray_DE0_w(:,:) => null (), Farray_DE0_r(:,:) => null ()
   real(ESMF_KIND_R8), pointer :: Farray_DE1_w(:,:) => null (), Farray_DE1_r(:,:) => null ()
+  real(ESMF_KIND_R8), pointer :: Farray_ug_DE0_w(:,:,:) => null(), Farray_ug_DE0_r(:,:,:) => null()
+  real(ESMF_KIND_R8), pointer :: Farray_ug_DE1_w(:,:,:) => null(), Farray_ug_DE1_r(:,:,:) => null()
 
   integer              :: rc, ncstat, ncid
   integer, allocatable :: computationalLBound(:),computationalUBound(:)
@@ -84,7 +87,7 @@ program ESMF_FieldIOUTest
   integer :: elem_tlb(1), elem_tub(1), elem_tc(1)
   integer :: tlb3(3), tub3(3), tlb4(3), tub4(3)
   integer :: i, j, t, endtime, k
-  logical :: failed
+  logical :: failed, allEqual
   real(ESMF_KIND_R8) :: Maxvalue, diff
 #if defined ESMF_NETCDF
   integer :: dim1id, dim2id, dim1len, dim2len, varid, ndims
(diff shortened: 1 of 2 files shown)


theurich:
@billsacks - Just a quick question with regards of this PR and 8.5.0. I see that this is work on `develop`, and that makes perfect sense. Also I do not feel any urgency to squeeze this into 8.5.0 (i.e. no user reports of this issue that we know of). But I want to confirm that this aligns with your thinking, too. Thanks.

billsacks:
Right - I do *not* intend for this to go into 8.5.0. Thanks for confirming.

theurich:
@billsacks - commenting on your point (1) from above. Your outline of steps taken does appear correct to me. Your hope to Redist() directly from `temp_array_p` to the original (user) provided Array wasn't even that far off. You can do this with ESMF, but then you do have to provide the `srcToDstTransposeMap` argument to allow mixing distributed and undistributed dims in a redist. Not very convenient here, especially because the information is already available within the arrays held by `temp_array_undist_p`.
The only suggestion I have perhaps wrt the code change is to move line 365, which sets `temp_array_undist_p` above the `if (has_undist) {` of line 364. Doing this will set `temp_array_undist_p` for all cases. I think then you can simply replace all of the use of `temp_array_p` with `temp_array_undist_p` in the entire block of line 383-417, and eliminate the extra `if (has_undist){` block 387-395. In my mind this follows more the style of the existing code where the code above the read() sets a number of pointer variables that can then be used consistently below, without further if-logic.

billsacks:
Thanks a lot for looking at this @theurich ! I like your suggestion; I'll make that change.

PR #178, opened 2023-08-14
Ensure the RPATH does not get blanked out during ESMX_App installation.

theurich:
This PR ensures that the RPATH that is correctly encoded during the ESMX_App build process is **not** blanked out during the _installation_ step. The CMake default _is_ to remove the encoded RPATH in executables during installation. I assume this is to support `LD_LIBRARY_PATH` setting during run-time.  However, for ESMF applications, and therefore ESMX, this is not a good idea. You typically encode the RPATH into the executable for good reason, so the correct shared library is found during loading without relying on `LD_LIBRARY_PATH`.

### DIFF
--- a/src/addon/ESMX/CMakeLists.txt
+++ b/src/addon/ESMX/CMakeLists.txt
@@ -41,6 +41,7 @@ endif()
 add_executable(${ESMX_EXE_NAME} ESMX_App.F90)
 target_include_directories(${ESMX_EXE_NAME} PUBLIC ${PROJECT_BINARY_DIR})
 target_link_libraries(${ESMX_EXE_NAME} PUBLIC esmx_driver)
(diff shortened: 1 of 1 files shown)


danrosen25 (approved):
This is fine with me. How did you encounter the issue? I'm brainstorming on the ESMX application prototypes and testing.

theurich:
I was working on a toy code that itself has a dependency on PnetCDF. I satisfied this dependency in the toy model's `CMakeLists.txt` file (since my local ESMF build was _not_ done with PnetCDF). The PnetCDF on my system is a shared library, and everything all the way up to linking the `build/esmx_app` was fine, setting the appropriate `RPATH` in the executable. However, during the _install_ step into `install/bin/esmx_app`, the RPATH was blanked out.
The result was that `build/esmx_app` could execute fine, while `install/bin/esmx_app` was not finding the PnetCDF library  during loading. 
The proposed change keeps the RPATH intact during for `install/bin/esmx_app`.

PR #179, opened 2023-08-14
Remove fixed GCC version from Spack build action

uturuncoglu:
This PR aims to make Spack Build action more flexible in terms of used gcc compiler. The new default is `gcc@latest`, which indicates that action will use latest available GCC compiler on runner. The user is also able to pass any specific gcc version like `gcc@11.4.0` etc. If specified compiler is missing, then action again uses latest available GCC compiler on runner.

### DIFF
--- a/.github/workflows/scripts/spack_concretize.sh
+++ b/.github/workflows/scripts/spack_concretize.sh
@@ -136,16 +136,25 @@ fi
 
 # check given gcc compiler is found or not? If not, use newer version
 if [[ "$comp" == *"gcc"* ]]; then
+  echo "::group::Check gcc compiler"
   comp_str=${comp/@/@=}
-  if [ -z "$(cat ~/.spack/linux/compilers.yaml | grep $comp_str)" ]; then
-    echo "Given compiler ($comp) not found! Try to find another ..."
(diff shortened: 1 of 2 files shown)


theurich (approved):
I am okay with the automatic switch to `gcc@latest` when a user specifies a version that cannot be found because of the output that indicates what is happening. Otherwise it might be surprising to the user why the requested version is ignored and a different one used. But with that output message, I think it is okay. Thank you!

uturuncoglu:
@theurich Do you want me to merge? In anytime, we could change the implementation and force to fail if the requested gcc version is not found.

theurich:
I do think failing with a good error message, if a specifically requested GCC version isn't found, is better. Especially because now the user can (and should) set `gcc@latest`, or just accept the default.

uturuncoglu:
@theurich Okay. Let me force to fail, if requested version is not found.

uturuncoglu:
@theurich Now it is failing. See following as an example, https://github.com/uturuncoglu/esmf/actions/runs/5861049405/job/15890408169

theurich (approved):
That looks good now. Thanks!!

theurich:
@uturuncoglu - please go ahead and merge into `develop`.

PR #181, opened 2023-08-24
Fix GitHub Actions for Deploying ESMF Docs

danrosen25:
Remove 3rd party GitHub Action
Uses deploy keys instead of personal access tokens
Separate deploy keys for ESMF_WEB_DEPLOY_KEY and ESMPY_WEB_DEPLOY_KEY are required
Code is always checking out and committing to default branches of 
- esmf-org.github.io
- esmpy_doc

Please use MERGE SQUASH

### DIFF
--- a/.github/workflows/build-esmf-docs.yml
+++ b/.github/workflows/build-esmf-docs.yml
@@ -42,12 +42,18 @@ jobs:
         name: esmf-docs
         path: ${{ github.workspace }}/artifacts
        
-    - name: Checkout esmf-org.github.io
-      uses: actions/checkout@v3
-      with:
-          repository: esmf-org/esmf-org.github.io
-          path: esmf-org.github.io
-          token: ${{ secrets.ESMF_WEB_TOKEN }}
+    - name: Setup deployment
+      run: | 
+        mkdir -p ~/.ssh
+        if [ -z "${{secrets.ESMF_WEB_DEPLOY_KEY}}" ]; then
+          echo "ERROR: Missing ESMF_WEB_DEPLOY_KEY!"
+          exit 1
+        fi
+        echo "${{secrets.ESMF_WEB_DEPLOY_KEY}}" > ~/.ssh/id_rsa
+        chmod 600 ~/.ssh/id_rsa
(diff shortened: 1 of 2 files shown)


billsacks (approved):
Thanks a lot for getting this working @danrosen25 ! I like that you are doing this with deploy keys and are avoiding using a 3rd party action. The approach that I was going to try was the one laid out in https://www.webfactory.de/blog/use-ssh-key-for-private-repositories-in-github-actions and the accepted answer in https://stackoverflow.com/questions/57612428/cloning-private-github-repository-within-organisation-in-actions/70283191#70283191 (involving `ssh-add - <<< "${{ secrets.SSH_PRIVATE_KEY }}"` rather than creating `~/.ssh/id_rsa`), but I don't know enough about this to know which solution is better in terms of security or if they are roughly equivalent. I'm happy to go with your solution.

danrosen25:
@billsacks Good find. It would be better to not write the private keys to disk. I can implement this change.

Explanation: What this does is adds the private key directly to the ssh-agent. In my code the private key is written to disk and then picked up by the ssh-agent. It's a vulnerability because (you probably already know this but) files aren't normally deleted from the disk when deleted from the file system. You'd have to have some high level access to the disk to run file recovery tools though.

theurich (approved):
Thank you for fixing the deployment implementation, and making it much safer. I reviewed the changes, but it is beyond my current level of understanding in this area.

I approve this on the grounds of seeing successful doc pushes to esmf-org.github.io from the deployment action from the PR branch fix/doc_deploy.

My assumption is that once this is merged into develop, it provides a functional and secure method to publish our online doc builds again.

danrosen25:
@billsacks @theurich 
I found a more elegant solution. It utilizes the github checkout action, which I trust more than a 3rd party application. I don't know why it's so hard to find this solution on the internet. GitHub doesn't seem to document it well, other than saying ssh-key is an input into github checkout.

This solution should make sense to everyone.

billsacks:
Cool, thanks for the latest change @danrosen25 - that indeed is a lot simpler.

PR #182, opened 2023-08-31
ESMX_LINK_PACKAGES to work with package namespaces and other minor clean-up

theurich:
This PR fixes an issue I encountered with the OpenMP and OpenACC packages under ESMX, using the `link_packages` feature under `application` in `esmxBuild.yaml`. Both these packages use namespaces, and cannot be used without namespace specification in `target_link_libraries()`. However, this call in `Driver/CMakeLists.txt` is redundant anyway, because if any component depends on a specific package (i.e. you need to use the `link_packages` option in the first place), then the component itself will have added the require `target_link_libraries()`. And the component will make that call with the correct namespace specification, e.g. for a specific language it uses, since locally it is known what it needs - after all it was the reason for the dependency in the first place.

All the other changes in this PR are minor clean-ups. A few typo corrections (in build options), bump ESMF dependency to 8.6.0, where it should be, and adding flags for the NVHPC compiler which I am currently experimenting with on Derecho.


danrosen25:
@theurich 
I see the issue. It's intended to give a way link targets to the esmx_driver at the driver level if they are shared across different applications (for a cleaner yaml file) but it doesn't have fine enough control for finding a package then linking the needed target. We should rethink how this works.

theurich:
@danrosen25 the approach in this PR works for what I am using the `link_packages` feature. It allows my ESMX applications to link against OpenMP and OpenACC packages correctly.
I'd like to merge the PR into develop just to move forward. Even if maybe we need to rethink some of the details down the road.

danrosen25 (approved):
Sounds good. And thanks for cleaning up some of my bugs.

PR #186, opened 2023-09-29
Feature/tclune/#185 eliminate cs global array

tclune:
This version "works".   It has a much smaller memory footprint.  Runs much faster and the results are more accurate by 3-4 digits.

But ... backward compatibility means we need to add a switch to preserve old behavior, and that will take a bit more work.

I wanted to get this in front of the core team before the gov't shutdown so that they might have advice to finish it up when I can return to work.   (Or maybe they'll take ownership from here?)

@atrayano is still in the process of verifying the memory reduction.


tclune:
I am thinking that the ultimate product here needs to be a separate module that can be identical in both dycore and in ESMF.   That way we can tell by inspection that the two codes are identical.     The module names will need to differ for practical reasons.

And probably safest to get this into dycore first, as they are the ones that must ultimately accept the numerical values.

tclune:
It looks like  `ESMF_UtilCreateCSCoords()` is not actually used.  Everything goes through `ESMF_UtilCreateCSCoordsPar()`.  If so, then this PR should delete the unused code.

Issue #184, opened 2023-09-27
Problems with I/O of Arrays with replicated dimensions


billsacks:
In doing some testing of #145 - specifically, testing the case that combines (a) multi-tiles, (b) more than 1 DE per PET, and (c) replicated dimensions - I noticed some oddities with I/O of Arrays with replicated dimensions. I have reproduced these oddities in a test case with single-tile Arrays: see https://github.com/billsacks/esmfprojects-replicated_dims_io (testing was done with billsacks/esmfprojects-replicated_dims_io@b96f2527520d43da452502b41248488d76a59713 and ESMF at the head of develop, 45807aad7a).

In a simple test case with 1 DE per PET, I get an output file that contains the replicated dimensions, but all slices of the replicated dimensions other than the first have garbage / missing values (see the attached output file `test.nc`). In a case with 2 DEs per PET, the output file does *not* contain the replicated dimensions: these are removed in the function `IO::redist_arraycreate1de` (see the attached output file `test_multiple_des_per_pet.nc`). Trying to read the file into a new Array then fails with this error:

```
20230927 090845.524 ERROR            PET1 ESMCI_Array.C:6272 ESMCI::Array::tRedistStore() Invalid argument  - srcArray and dstArray must provide identical number of exclusive elements
20230927 090845.524 ERROR            PET1 ESMCI_Array.C:6052 ESMCI::Array::redistStore() Invalid argument  - Internal subroutine call returned Error
20230927 090845.524 ERROR            PET1 ESMCI_IO.C:396 ESMCI::IO::read() Invalid argument  - Internal subroutine call returned Error
20230927 090845.524 ERROR            PET1 ESMCI_IO.C:282 ESMCI::IO::read() Invalid argument  - Internal subroutine call returned Error
20230927 090845.524 ERROR            PET1 ESMCI_Array.C:3970 ESMCI::Array::read() Unable to read from file  - Internal subroutine call returned Error
20230927 090845.524 ERROR            PET1 ESMCI_Array_F.C:833 c_esmc_arrayread() Unable to read from file  - Internal subroutine call returned Error
20230927 090845.524 ERROR            PET1 ESMF_ArrayHa.F90:713 ESMF_ArrayRead Unable to read from file  - Internal subroutine call returned Error
20230927 090845.524 ERROR            PET1 replicatedDimsIO.F90:74 Unable to read from file  - Passing error in return code
```

So the problems I see are:
[...]

PR #187, opened 2023-10-10
Add replicatedDimCount to ESMF_ArrayGet, and improve some unit tests of replicated dimensions

billsacks:
This started with my work towards addressing esmf-org/esmf#184 . That turned out to be a bigger problem than I anticipated, but some of the work towards addressing that issue seemed worth preserving.

This PR:

(1) Introduces a method to get the replicated dimension count of an array and uses this method in place of some previously-duplicated code (in both ESMCI_Array.C and ESMCI_IO.C).

(2) Adds a replicatedDimCount optional output argument to ESMF_ArrayGet, with associated unit tests.

(3) Improves unit tests of Arrays with replicated dimensions in ESMF_ArrayIOUTest.F90 - most notably, outputting the results of different tests to different files to aid later examination, and making the values in the Array vary within each PET to make the tests more rigorous.


src/Infrastructure/Array/include/ESMCI_Array.h
```diff
@@ -385,6 +385,8 @@ namespace ESMCI {
     const int *getTotalLBound()             const {return totalLBound;}
     const int *getTotalUBound()             const {return totalUBound;}
     int getTensorCount()                    const {return tensorCount;}
+    int getReplicatedDimCount()             const
```
billsacks, line 388: Note that I don't have a return code (rc) for this function, since nothing can currently lead to it failing. I was thinking of this as just a small step beyond a simple getter. Does that seem reasonable, or should this have a place-holder rc?

src/Infrastructure/Array/interface/ESMF_ArrayGet.cppF90
```diff
@@ -129,7 +129,7 @@ contains
 ! !INTERFACE:
   ! Private name; call using ESMF_ArrayGet()
   subroutine ESMF_ArrayGetDefault(array, keywordEnforcer, arrayspec, typekind, &
-    rank, localarrayList, indexflag, distgridToArrayMap, &
+    rank, replicatedDimCount, localarrayList, indexflag, distgridToArrayMap, &
```
billsacks, line 132: Does this seem like a good place to insert this new optional output argument? Another possible position is after `dimCount`, though the stuff there seems related solely to the `distGrid`, so that didn't feel quite as appropriate.

billsacks (commented):
@theurich - see a couple of inline questions I have about this PR.

Also, note that I ended up removing the C++ unit test class (db668ee7658e09bd19a5330a241bbe235053b095): its tests were completely redundant with the new Fortran-level tests, so it seemed like more of a burden than a help to have this.

src/Infrastructure/Array/src/ESMCI_Array.C
```diff
@@ -650,9 +650,7 @@ Array *Array::create(
     }
   }
   // determine replicatorCount
-  int replicatorCount = 0;  // initialize
-  for (int i=0; i<dimCount; i++)
-    if (distgridToArrayMapArray[i] == 0) ++replicatorCount;
+  int replicatorCount = getReplicatedDimCountImpl(dimCount, distgridToArrayMapArray);
```
theurich, line 653: Change to `getReplicatedDimCount()`.

src/Infrastructure/Array/src/ESMCI_Array.C
```diff
@@ -1400,9 +1398,7 @@ Array *Array::create(
...
```
theurich, line 1401: Change to `getReplicatedDimCount()`.

billsacks:
@theurich - thanks a lot for your comments here. I only have a minute right now, so have just skimmed them for now, but I wanted to explain the need (I think) for `getReplicatedDimCountImpl`: this is called from `Array::create` as well as from the public `getReplicatedDimCount`. The call from `Array::create` is done before the Array object is constructed. This led me to think that I should create a static method to do the work, at least if I wanted to remove the duplicated code from all places. Does that make sense, or do you still think I should remove this Impl method? It's quite possible I'm missing something here.

theurich:
Yes, being functional inside `Array::create()` is an issue. My preference in this case would be to implement just a local utility routine to cover that phase of Array construction. Public facing, both on the ESMCI level as well as the ESMF access is through the instance method. But let's discuss tomorrow on the core call.

theurich:
And on second thought, it is really a DistGrid method. Still fine to route it through the ESMCI::Array and ESMF_Array API for convenience (as you have it), but I think the implementation should be an instance method of ESMCI::DistGrid.

billsacks:
@theurich - I made your suggested change to the placement of the `replicatedDimCount` argument.

Yes, let's talk tomorrow about your other comments. I thought that it didn't make sense to put this new function in the DistGrid class because it's mainly working with `distgridToArrayMap`, which I think is specific to the Array, not the DistGrid. I'm happy to make this a helper function rather than a static method (I see this is done elsewhere in ESMCI_Array.C, e.g., `sparseMatMulStoreNbVectors`), but I don't understand your whole vision, e.g., making this helper method public-facing. Anyway, let's talk more tomorrow so I can better understand your thinking here.

theurich:
Yes, the DistGrid method was a dumb idea, with half my brain out the door already yesterday afternoon. Just ignore it please.
My thinking this morning - hopefully with more brain engaged this time - is to add a member variable `replicatedDimCount` in the Array class. It would be set during `Array::create()`. Probably still worth having an internal utility routine that implements it, since used in more than one place. But not worth a static class method. Finally the public accessor `getReplicatedDimCount()` is simply returning the `replicatedDimCount` member.

billsacks:
I have checked the new [API documentation](https://earthsystemmodeling.org/docs/nightly/replicated_dims_cleanup/ESMF_refdoc/node5.html#SECTION050551800000000000000) and confirmed that it looks good.

billsacks:
@theurich - I think I have made the changes you suggested (making a member variable for number of replicated dimensions); can you look this over and let me know if it looks good to you now?

However, I didn't see where / how to change the calculation of buffer size in the serialize / deserialize. If that's needed, can you point me to what I need to do for it? (Unit tests currently pass, so maybe nothing more is needed?)

theurich (changes requested):
This all looks good, except for one small change needed (answering your question about the serialize buffer size inquire part).
Code at https://github.com/esmf-org/esmf/blob/13ac9cacdd845d5dbf073769cb081ad86fc310df/src/Infrastructure/Array/src/ESMCI_Array.C#L4468
```
ip += 2 + 2*tensorCount + 2*distgrid->getDimCount () +
      rank + delayout->getDeCount ();
```
needs to be changed to
```
ip += 3 + 2*tensorCount + 2*distgrid->getDimCount () +
      rank + delayout->getDeCount ();
```
i.e. bump up the first number by one, because there is one more `int` member now.

billsacks:
Thanks @theurich ! I made that change, reran local testing (all good) and have pushed it. Once you approve, I'll merge this.

theurich (approved):
Bill, this looks good to me now. Please merge at your convenience. Thanks!

PR #194, opened 2023-11-10
Add artifacts to Spack build action for debugging

uturuncoglu:
This PR aims to add new feature to Spack build action to store run directory of NUOPC App Prototypes as artifacts.


uturuncoglu:
@theurich Here is the PR to add artifacts to Spack build action. Please let me know if you need any change. At this point, if any build in the matrix fails, the action stop completely and only artifacts of the failed job appears.

theurich:
@uturuncoglu  I merged this PR into my own branch for testing... and because of that I cannot re-open the PR. I actually do have a request for change before this goes into develop, but now I cannot re-open the PR. Could you please create a new PR for this when you get back? Thank you!

PR #196, opened 2023-11-17
Add an env var to allow skipping performance tests

billsacks:
Add a new environment variable ESMF_TESTPERFORMANCE. If set to "OFF", this will skip checking performance in performance tests.

This is useful on machines where performance tests are highly variable and can lead to spurious failures. e.g., this happens on green (a MacOS / Darwin system), where the nightly cron job can run very slowly, leading to failures in some performance tests (generally in ESMF_ArrayRedistPerfUTest).

See https://github.com/esmf-org/esmf-test-scripts/commit/229f023a9e1ad25a40715be23272c110efc78b82 for how I am using this for nightly testing.


billsacks:
@theurich can you please give this a quick look whenever you get a chance and let me know if this implementation is in line with what you imagined?

theurich (approved):
@billsacks  That looks exactly right to me!

PR #199, opened 2023-11-28
Simplify uploaded artifacts

theurich:
Simplifications:
- Only capture what is collected under the `NUOPC-PROTO-RESULTS` directory when running the NUOPC protos.
- Rely exclusively on the `actions/upload-artifact@v3` compression (zip), rather than tar and gzip. This simplifies artifacts usage.

### DIFF
--- a/.github/workflows/test-build-spack.yml
+++ b/.github/workflows/test-build-spack.yml
@@ -134,12 +134,19 @@ jobs:
           -r ${{ github.workspace }}/nuopc-app-prototypes \
           -s ~/.spack-ci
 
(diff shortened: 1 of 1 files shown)


theurich:
@uturuncoglu This PR goes into your branch. If you are okay to pull this in, then we can bring your PR into develop. Just was simpler because I wanted to test this out under actions, so I needed a branch to work with for that.

uturuncoglu (approved):

uturuncoglu:
@theurich It is done.

PR #201, opened 2023-12-05
Get ESMF Compiling on Cygwin, and set up a CI run

DWesl:
I updated my ESMF build on Cygwin, and tried to generalize my changes from the last time I did this so ESMF would still compile on Linux.

I also added a CI run to ensure ESMF compilation gets checked.  I think I set up the tests to run as well, but I'm less sure of that.

Might close #136; I'm not sure if they're after a Cygwin build of ESMF or a 64-bit MinGW/Windows build of ESMF.  PR #149 suggests they're after a MinGW-w64 build but are trying to do it in Cygwin without cross-compilers for whatever reason.

PR #202, opened 2023-12-08
BLD: Get ESMF compiling on Cygwin

DWesl:
I updated my ESMF build on Cygwin, and tried to generalize my changes from the last time I did this so ESMF would still compile on Linux.

I also added a CI run to ensure ESMF compilation gets checked. I think I set up the tests to run as well, but I'm less sure of that.

Might close https://github.com/esmf-org/esmf/issues/136; I'm not sure if they're after a Cygwin build of ESMF or a 64-bit MinGW/Windows build of ESMF. PR https://github.com/esmf-org/esmf/pull/149 suggests they're after a MinGW-w64 build but are trying to do it in Cygwin without cross-compilers for whatever reason.

Follow-on to #201.

### DIFF
--- a/.github/workflows/test-build-cygwin.yml
+++ b/.github/workflows/test-build-cygwin.yml
@@ -1,0 +1,74 @@
+name: Test on Cygwin
+on: ["push", "pull_request"]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+env:
+  SHELLOPTS: igncr
+  CYGWIN_NOWINPATH: 1
+  CHERE_INVOKING: 1
+  ESMF_DIR: /cygdrive/d/a/esmf/esmf
+  ESMF_BOPT: O
+  ESMF_OPTLEVEL: 2
+  ESMF_COMM: mpiuni
+  ESMF_COMPILER: gfortran
+  ESMF_TESTSHAREDOBJ: ON
+  ESMF_RANLIB: ranlib
+  ESMF_LAPACK: system
+  ESMF_LAPACK_LIBPATH: /usr/lib
+  ESMF_LAPACK_LIBS: -llapack -lblas
+  NETCDF: /usr
+  ESMF_NETCDF: split
+  ESMF_NETCDF_INCLUDE: /usr/include
+  ESMF_NETCDF_LIBPATH: /usr/lib
+  ESMF_CXX: g++
+  ESMF_CXXCOMPILEOPTS: "-g -O2"
+  ESMF_F90: gfortran
+  ESMF_F90COMPILEOPTS: "-g -O2"
+  ESMF_INSTALL_PREFIX: /tmp/fakeroot/usr
+  ESMF_INSTALL_DOCDIR: share/doc/esmf
+
+jobs:
+  cygwin_build_test:
+    runs-on: windows-latest
+    steps:
+      - run: git config --global core.autocrlf input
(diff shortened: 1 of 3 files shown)


theurich:
@oehmke, @billsacks, @danrosen25, @uturuncoglu, I am not sure what's the best move forward here. The branch builds again for me under Linux after fixing some syntax issue in the logic change. However, I have not looked at the action part of the PR, and will not have time for it either. @oehmke I assigned the PR to you just to have a person responsible for it, and made the rest of the core team reviewers. Hope this works for now.

billsacks (approved):
@theurich if you're happy with the changes to common.mk and the Cygwin build_rules.mk then I am, too. I'll defer to others who have more experience with github workflows to review the workflow piece.

.github/workflows/test-build-cygwin.yml
```diff
@@ -0,0 +1,74 @@
+name: Test on Cygwin
+on: ["push", "pull_request"]
```
danrosen25, line 2: @DWesl
Although we like the idea of CI testing all features we aren't ready for this style. I suggest changing this to
`on: ["workflow_dispatch", "pull_request", "release"]`

DWesl: Done.

DWesl:
Presumably the doc build is tested other places, so I skipped that to lower the compilation time and avoid the failure.

The ESMPy tests are also gone skipped now, I think those were failing and I couldn't figure out why.

At the moment, the action is checking that ESMF and ESMPy build on Cygwin and the ESMF checks pass on Cygwin, and will do that anytime someone opens a PR, pushes commits to a PR, pushes a button to run the action, or hits the "Release" button on GitHub.

If you'd prefer not to spend three hours waiting for all that to run on every PR, I can drop the action now that it's demonstrated the rest of the changes allow Cygwin builds, and leave just the Cygwin changes.

DWesl:
Should I drop the GHA CI to just `workflow_dispatch`, so the 3-hour Cygwin build only runs when someone specifically requests it?

danrosen25:
@DWesl 
Does it take 180 minutes on your local machine too? I would try to integrate Cygwin testing into our CI but I'm trying to keep it under 30 minutes. What part of the workflow was spent in setting up the Windows machine versus running `make` and `make check`?

We should try to automate this so that it runs before we finalize a release.

DWesl:
I don't remember how long it took last time I built ESMF locally, but it wasn't much shorter if it was at all.  It's deleted the logs from the last run, but I would expect the Windows/Cygwin set-up took around ten minutes, then over an hour each for `make` and `make check`.  Should I push the button to run the workflow again and see how long each step takes?

I couldn't work out from [the documentation](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release) how to get it to run before a release, but I think I told it to run on pre-releases, does that work?

danrosen25:
@DWesl 
Agreed, it's not clear if `{on: {release: {types: [unpublished]}}}` or `{on: {create: {ref_type: [tag]}}}` would work. I don't think they will work. But we could test every release tag (including beta releases) with the following.

```
on:
  push:
    tags:
      - 'v[0-9]+.[0-9]+.[0-9]+'
```

DWesl:
The Cygwin install takes about seven minutes, then the build takes about half an hour to run into the PIO issues from #348 and crash.

DWesl:
With the PIO fixes merged, Cygwin takes about nine minutes to install dependencies and two and a half hours to build ESMF.  Building ESMF once that's done takes eleven seconds, then the ESMF check takes another hour.

PR #204, opened 2023-12-23
ESMPY: Add ability to mask nodes of a Mesh

mcflugen:
This pull request adds the ability for a user to associate a node mask with an ESMPY Mesh.

The *Mesh.add_nodes* method now accepts a keyword, *node_mask*, through which a user can pass a mask array of length *n_nodes*. Previously it was only possible to associate a mask with a mesh's elements. This new functionality follows the same pattern as *Mesh.add_elements*.

To make this work I had to add an extra argument, *nodeMask*, to *ESMC_MeshAddNodes* and, as such, modified all existing calls to this function to include the value *NULL* for this extra argument (i.e. no masking).

All of the existing tests pass and this fixes the issue I was having, but I haven't added any new tests yet but I am happy to do so.

This, of course, supersedes #203, which was created in error.


mcflugen:
@oehmke This is the PR we talked about, if you want to have a look when you find some time.

mcflugen:
To do,
* [ ] Add tests
* [ ]  Update docs

mcflugen:
I think this is ready for someone to have a look at now.

oehmke (commented):
This looks good to me. I made one small change so to use an unallocated structure, so we don't have to deallocate it later. I'm going to see if Ryan is available to also review, otherwise looks good to me.

rokuingh:
@mcflugen Thanks for this contribution! I am just getting back into the swing of things around here. I spent what time I have today, and noticed several esmpy test failures with this pr. Didn't have much time to dig, but it looks like the internal Mesh C++ layer is receiving invalid node_ids: 

> 20240207 122424.785 ERROR            PET0 /home/ryan/Dropbox/sandbox/esmf/src/Infrastructure/Mesh/src/ESMCI_Mesh_Glue.C:189 ESMCI_meshaddnodes() Value unrecognized or out of range  -  node ids must be >= 1

This error came from running the following test:

`python3 -m pytest src/esmpy/test/test_api/test_mesh.py::test_add_nodes_with_mask[None]`

I have a small amount of time in the mornings right now, so I'll keep digging when I can.

rokuingh:
I just pushed some mods to fix the esmpy test failures on this pr. There is one failure remaining, although it seems unrelated at the moment, will check on develop.

oehmke:
What’s the failure?

billsacks:
I just tried running ESMPy testing on this branch (updated to the latest version of develop) and got failures in the newly-added tests. (If I backed out the changes to the two test files in the PR, all ESMPy tests passed.)

@rokuingh you mentioned having pushed fixes for the tests, but I don't see any commits in this PR from you. Is it possible you pushed the changes to somewhere else? When you get a chance can you please push them to this branch or point us to the branch with the fixes, then we can confirm that all tests pass before merging?

anntsay:
Waiting on Ryan's response.. will make deicsion on July 3rd based on whether we hear from Ryan.

billsacks:
@rokuingh pointed me to the branch where he has fixed the test issues: https://github.com/esmf-org/esmf/tree/pr202 - specifically see https://github.com/esmf-org/esmf/commit/dc53aad6082be6f7f09dc9737934d251d59510c1 . I am confirming that tests pass on that branch; once I do, I'll merge it.

billsacks:
The latest commit from @rokuingh fixes the ESMF testing, but the ESMPy testing is still failing for me as noted above.

billsacks:
Based on an email exchange with @rokuingh , I will fix the ESMPy testing issues. It seems like the main issue is that the new regridding tests only work in serial, so I will add that limitation. While looking into that, though, I began to feel like we could reduce the number of new tests in this PR and still keep good coverage of the new features. I'm going to try to do that: While both @rokuingh and I both really (hugely!) appreciate the addition of good test coverage of these new features, I want to try to avoid having too much of a disproportionate number of tests for this, and I particularly want to avoid having tests with random inputs if possible (since they can be hard to diagnose if they fail).

billsacks:
I also thought I'd try to move the new test functions into their respective test classes, but it looks like that doesn't work with parametrized functions, which I guess is why they were written outside of the classes to begin with.

billsacks:
My original thinking was to reduce the regridding tests considerably, to end up with only 2 regidding tests:
- I'm thinking we don't need to cover the case with node_mask=None, because I assume that's already implicitly covered in pre-existing tests
- I think we could keep a single test of test_regrid_with_node_mask using one of the five current masks and a single regridding method
- I would remove test_regrid_with_random_node_mask, because I don't like randomness in tests: it can be problematic when you get a test failure and then can't reproduce it later. It does look like this test differs a bit from test_regrid_with_node_mask, in the destination coordinates in this random one differ from the source coordinates, whereas the destination and source are identical in the non-random version (if I understand the code correctly)... but I think that, for the purpose of what this is trying to test, that shouldn't matter.
- I would keep a single test of test_regrid_with_multivalued_node_mask, using a single regridding method. I would reduce the size considerably – e.g., a 3x4 array rather than 100x200 – and use a hard-coded array for src_data rather than the current random values

But @rokuingh pointed out that it can be good to have some extra test coverage of regridding via ESMPy, so supported keeping more tests in place. So I'm doing a more limited reduction:
- I'm removing `test_regrid_with_random_node_mask`, to remove this test with randomness and because it doesn't look to me like it adds significant coverage
- I'm changing `test_regrid_with_multivalued_node_mask` to use a small, hard-coded source array rather than a random array, both to remove randomness and to reduce the test time. (With this change, the added test time from this PR is insignificant, whereas previously, the tests in this PR added about 9 seconds, I _think_ mainly due to this larger test.)

I considered also changing test_regrid_with_node_mask to use a single mask... I tried creating a mask that felt like a hybrid of the existing masks, `[[1, 0, 0, 1], [0, 1, 1, 0], [1, 1, 0, 0]]`, but bilinear and patch remapping failed with that mask. This made me think that I don't understand this well enough to choose a single mask for this test, so I left this parametrized test as is.

billsacks:
The test changes mentioned above are in https://github.com/esmf-org/esmf/pull/204/commits/3fbcbd9a3bde0d46054d82bd42f69ec092002352 and https://github.com/esmf-org/esmf/pull/204/commits/8bfcbee121aba59ed4a159ee29a5df1ae7f61095. @mcflugen if you feel that I removed something important by doing this, please let me know. I plan to merge this shortly, but can bring back these removed tests if it seems important.

Thank you very much for this very useful contribution @mcflugen ! We really appreciate your work on this!

PR #205, opened 2024-01-05
Enable PIO with mpiuni

### LINKED ISSUE #131: Allow PIO builds with mpiuni
Currently, builds with PIO can only be done with a real mpi library, not mpiuni. This can cause problems for people using ESMPy, since ESMPy is sometimes built without mpi – and this is apparently needed on many HPC systems (see also https://github.com/conda-forge/esmpy-feedstock/issues/70). @oehmke mentioned that others have stated a desire for this – e.g., to use the offline RegridWeightGen.

From last week's core team meeting: Getting this working would involve:
- We would need to figure out how to point pio to mpiuni. It's probably relatively little effort to get this working for the internal pio. But it gets tricky with an external pio: would need to build pio with a mpiuni library, and need to make sure esmf is using the same thing as pio.
- Making sure that pio works with our mpiuni library; if not, changing it or switching to a different mpiuni library.

The general feeling was that a reasonable compromise would be to support mpiuni with internal PIO, but not when using an external PIO library.

Note that if/when this is done, we should back out the comment about needing an mpi library for the PIO build that I introduced in https://github.com/esmf-org/esmf/pull/130

billsacks:
This PR enables building and running with the internal PIO when using mpiuni. This is especially relevant for people using ESMPy, since ESMPy is sometimes built without mpi – and this is apparently needed on many HPC systems (see also https://github.com/conda-forge/esmpy-feedstock/issues/70). This resolves #131 .

I still want to do some more testing of this and look into whether there are any tests that are currently disabled with mpiuni that should be enabled. But I have done significant testing with ESMPy and have verified that we can now run ESMPy's test suite with mpiuni and all tests pass in that configuration. So it feels like this is getting close enough that it's worth opening a PR.

@theurich and/or @oehmke - I would welcome a review of this. I am assigning both of you as reviewers but will let you decide if you have the time and interest to review it. I will make some line comments calling out pieces that I think most warrant a careful double-check.


src/Infrastructure/stubs/mpiuni/mpi.c
```diff
@@ -179,6 +199,84 @@ int Petsc_MPI_Finalize(void)
...
+  //
+  // From some experimentation with a real MPI library, the MPI_Type_size of newtype is
+  // independent of the value of stride. Since the MPI_Datatype in mpiuni is just the size
+  // of the datatype, we ignore the possible complexity of stride in this implementation.
+  *newtype = count*blocklength*oldtype;
+  return MPI_SUCCESS;
+}
+
+int ESMC_MPI_Type_create_indexed_block(int count, int blocklength,
+                                       const int array_of_displacements[],
```
billsacks, line 249: Note that I ignore array_of_displacements here: it seems like, for mpiuni, the important thing about type creation is to end up with a value that reflects the size of the resulting type, and from some experimentation with this MPI function with openmpi, it seems like the MPI_Type_size of the result is independent of the displacements. (I originally had a check that these displacements were evenly spaced, but removed this check (in commit 59a996da1f) because it seems unnecessary if what we care about is the size of the result.)

But I would be happy to have a second set of eyes on this.

src/Infrastructure/stubs/mpiuni/mpi.c
```diff
@@ -179,6 +199,84 @@ int Petsc_MPI_Finalize(void)
...
+  // Displacements are not implemented so return an error code if they are non-zero
+  if (displs[0] != 0) {
+    return MPI_ERR_INTERN;
+  }
+
+  MPIUNI_Memcpy(recvbuf, sendbuf, sendcounts[0]*sendtype, CHECK_FOR_MPI_IN_PLACE_DEST);
+  return MPI_SUCCESS;
+}
+
+int ESMC_MPI_Type_create_hvector(int count, int blocklength, MPI_Aint stride,
```
billsacks, line 236: Note that I ignore stride here: it seems like, for mpiuni, the important thing about type creation is to end up with a value that reflcts the size of the resulting type, and from some experimentation with this MPI function with openmpi, it seems like the MPI_Type_size of the result is independent of the stride.

However, this function is the one that I'm least certain of, so would be happy to have a second set of eyes on this.

oehmke:
Sorry for not responding sooner. I'll review this soon (hopefully this week), and we can check if Gerhard also wants to take a look.

billsacks:
No rush! As you may see from the commit dates, this has been very on-again-off-again work anyway. And I'm actually thinking of putting this on pause so I can work on prototyping the Julia model coupling next, since that feels high priority in terms of time.

billsacks:
I have run testing on my Mac on the latest version (cc470b7362), both with openmpi and mpiuni. All tests pass (using exhaustive testing), and ESMPy tests pass.

oehmke (approved):
Looks good to me. Thanks for all of your hard work going down all the rabbit holes on this one!

billsacks:
I have done enough testing on this that it feels ready to merge, and I'm going to do so in order to get nightly test results.

I still want to add some more unit tests that invoke I/O and are run with mpiuni: I noticed that much of our I/O testing (e.g., ArrayIOUTest and similar) is only set up for multi-processor testing. My thought is to add at least some basic Array I/O tests that are run when testing with mpiuni. I'm thinking that I'll do that by introducing a new test file like ESMF_ArrayIOBasicUTest, where the Array setup will work on single-processor testing. But I want to get this merged to see initial test results, so I'll add that later. @oehmke let me know if you have any thoughts on this plan of adding a new unit test file for the sake of adding some tests that will work with mpiuni: it feels a little weird to me to add a new file for that purpose, but I don't see a way around it other than doing some potentially significant rework of all of the tests in ESMF_ArrayIOUTest to support single-processor testing - since my understanding is that, if any tests in a file only support multi-processor, then ALL tests in the file need to be for multi-processor only.

Issue #192, opened 2023-10-30
Bug in Config layer


tclune:
[reproducer_192.tgz](https://github.com/esmf-org/esmf/files/13925211/reproducer_192.tgz)

The following program produces incorrect results when run using the resource file below.
```f90
program main
  use ESMF
  implicit none
  type(ESMF_VM) :: vm
  type (ESMF_Config) :: cf
  character*100 :: fname
  logical :: ispresent
  character(len=ESMF_MAXSTR) :: string
  integer :: status

  call ESMF_Initialize(vm=vm, rc=status)

  fname = 'test_config.rc'
  cf = ESMF_ConfigCreate(rc=status)
  call ESMF_ConfigLoadFile( cf, fname, unique = .true., rc=status)


  call ESMF_ConfigFindLabel(cf, 'CB',  isPresent=ispresent, rc=status)
  print*, '----'
  print*, 'bf configset-A,  CB: ispresent=', ispresent

  string='A1XXX'
  call ESMF_ConfigSetAttribute( cf, trim(string), label='CA:', rc=status)

  call ESMF_ConfigFindLabel(cf, 'CB',  isPresent=ispresent, rc=status)
  print*, '----'
  print*, 'af configset-A,  CB: ispresent=', ispresent

end program main
```
`test_config.rc:`:
```
CA: 'A1'
CB: 'B1'
```

tclune:
Found output:
```
 /Users/tclune/Fortran/ESMF_config_bug/main.F90          19 T           0
 /Users/tclune/Fortran/ESMF_config_bug/main.F90          23           0
 /Users/tclune/Fortran/ESMF_config_bug/main.F90          26 F           0
```


A clue is that the reproducer depends on the new value being a longer string than the original from the rc file.

theurich:
Likely not compiler dependent, but just in case, Tom says that he ran the reproducer on NAG and gfortran-13 to see it fail.

tclune:
To clarify:  I've only had the failure with gfortran 12.3.  NAG succeeds.  @mathomp4 has been unable to replicate on his nigh identical laptop.   We are mystified.

tclune:
Sorry I confused this ticket with my open issue on "Info" (not Config).

tclune:
Understood.   Really too bad that it cannot be just jetisoned in favor of YAML.    Backward compatibility and all that. :-)

danrosen25:
[...]

PR #207, opened 2024-01-11
Bug fixes for ESMF_Config

danrosen25:
Fixes [esmf-support 184](https://github.com/esmf-org/esmf-support/issues/184)
Fixes [esmf-support 155](https://github.com/esmf-org/esmf-support/issues/155)
Fixes [esmf 192](https://github.com/esmf-org/esmf/issues/192) 

* Fix issue with inserting characters to config buffer
* Fix issue replacing attributes that start with quotations
* Removes line limits from ESMF Config
* Remove this_line from ESMF_Config
* Add next_item index marker to iterate over lists
* Add eolflag marker to iterate over lists
* Set value to first non colon after find label to support labels with and without colon
* Fix blank values
* Replace ESMF_ConfigSetIntI4 internals with ESMF_ConfigSetString for less maintenance
* Increase value size limit in multiple subroutines to NBUF_MAX

### DIFF
--- a/src/Infrastructure/Config/src/ESMF_Config.F90
+++ b/src/Infrastructure/Config/src/ESMF_Config.F90
@@ -199,6 +199,10 @@
                                                   ! will do a bound check
 
        character, parameter :: BLK = achar(32)   ! blank (space)
+       character, parameter :: QTD = achar(34)   ! double quotation "
+       character, parameter :: CMT = achar(35)   ! number sign #
+       character, parameter :: DSN = achar(36)   ! dollar sign $
+       character, parameter :: QTS = achar(39)   ! single quotation '
        character, parameter :: TAB = achar(09)   ! TAB
        character, parameter :: EOL = achar(10)   ! end of line mark (newline)
        character, parameter :: EOB = achar(00)   ! end of buffer mark (null)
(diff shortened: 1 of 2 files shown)


danrosen25:
Fixes issue #192

danrosen25:
@oehmke to kick off all ESMF tests on derecho for branch bugfix/esmf_config

oehmke (approved):
I looked through this and overall it looks good! One question that I had though was on line 3372. I wasn't sure why there was a -1. Would you take a look at that and make sure that it's ok? Other than that, I'm good with this going in.

billsacks:
Thanks for inviting me to review this. I don't feel a need to review it, but can do so if you feel it would be worthwhile. Let me know your feelings.

src/Infrastructure/Config/src/ESMF_Config.F90
```diff
@@ -3283,15 +3367,20 @@ subroutine ESMF_ConfigParseAttributes( config, unique, rc )
       do while ( i .lt. config%cptr%nbuf )
 
         ! get next line from buffer
-        j = i + index_(config%cptr%buffer(i:config%cptr%nbuf), EOL) - 1
-        this_line = config%cptr%buffer(i:j)
+        j = index_(config%cptr%buffer(i:config%cptr%nbuf), EOL)
```
danrosen25, line 3370: @oehmke 
`i` is the beginning index of a character string
`j` is the ending index of a character string.
`index_` searches for the next EOL (end of line) character.
`%nbuf` is the used buffer space, the last character in used buffer space is supposed to be a EOB (end of buffer) character

if EOL is found then this_label is everything from `i` to `EOL`, inclusive. This matches the old code but the old code assumed that you would always find EOL, which in theory is true.
if a EOL isn't found then this_label is everything from `i` to `EOB - 1` (so excluding the EOB character) and previously it would actually be `i` to `i - 1`, which would be an undefined string.

oehmke: Ah, so the -1 is to avoid the EOB. Looking at the case where it finds the EOL, then the max of that would be %nbuf-1 too, because the last character is EOB not EOL. Ok, that all makes sense to me now. Nice work. I'm fine with you merging this now.

PR #221, opened 2024-02-02
Fix build of libesmftrace_preload.dylib on Darwin-openmpi

billsacks:
With Darwin-openmpi, we use a Fortran linker for libesmftrace_preload.dylib. But the link command was using flags appropriate for a C++ linker rather than a Fortran linker. This PR changes the link flags to be appropriate for a Fortran linker in this case.

I have tested this on my Mac (green, with gfortranclang+openmpi) and confirmed that this resolves the build issues I was running into.

I have also tested this on catania and derecho to ensure that this doesn't break the build / test on other systems.

@theurich I'd welcome a quick look.


theurich (approved):
This looks really nice and clean. Thank you @billsacks for cleaning up this mess!

PR #223, opened 2024-02-03
Add 'ESMF::ESMF' CMake target alias to support unambiguous linking

AlexanderRichert-NOAA:
This PR adds an alias to the CMake target `ESMF` called `ESMF::ESMF`. This way, when a downstream package calls `target_link_libraries(ESMF::ESMF)` it will fail if the target isn't found, rather than generating a spurious and unwanted `-lESMF` flag. Using the alias leaves the `ESMF` target itself untouched, so no adverse consequences are anticipated.

### DIFF
--- a/cmake/FindESMF.cmake
+++ b/cmake/FindESMF.cmake
@@ -109,6 +109,9 @@ if(EXISTS ${ESMFMKFILE})
     endif()
   endif()
 
+  # Add target alias to facilitate unambiguous linking
+  add_library(ESMF::ESMF ALIAS ESMF)
+
   # Add ESMF include directories
   set(ESMF_INCLUDE_DIRECTORIES "")
   separate_arguments(_ESMF_F90COMPILEPATHS UNIX_COMMAND ${ESMF_F90COMPILEPATHS})
--- a/cmake/FindESMF.cmake
+++ b/cmake/FindESMF.cmake
@@ -110,7 +110,9 @@ if(EXISTS ${ESMFMKFILE})
   endif()
 
   # Add target alias to facilitate unambiguous linking
-  add_library(ESMF::ESMF ALIAS ESMF)
+  if(NOT TARGET ESMF::ESMF)
+    add_library(ESMF::ESMF ALIAS ESMF)
+  endif()
 
   # Add ESMF include directories
   set(ESMF_INCLUDE_DIRECTORIES "")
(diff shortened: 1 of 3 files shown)


AlexanderRichert-NOAA:
@theurich

theurich:
@danrosen25 Assigning to you to make final decision on pulling into `develop`. It looks good to me.

theurich (approved):
@AlexanderRichert-NOAA looks good to me. Thanks!

theurich:
@danrosen25 @oehmke I suggest considering this change also for `patch/8.6.1`.

oehmke:
Makes sense. Thanks for taking a look.

danrosen25:
@oehmke @theurich 
I'll have to cherry pick 9f0a8be064b1793c9bef5cf16d05dafee7a96dd5 into 8.6.1 but I can do this. Let me run a quick test and make sure we capture all of the cmake modules we provide.

theurich:
> @oehmke @theurich I'll have to cherry pick [9f0a8be](https://github.com/esmf-org/esmf/commit/9f0a8be064b1793c9bef5cf16d05dafee7a96dd5) into 8.6.1 but I can do this. Let me run a quick test and make sure we capture all of the cmake modules we provide.

It should go into both `develop`, and `patch/8.6.1`. My suggestion would be PR merge into `develop`. Then cherry-pick the hash from `develop` into `patch/8.6.1`.

cmake/FindESMF.cmake
```diff
@@ -109,6 +109,9 @@ if(EXISTS ${ESMFMKFILE})
     endif()
   endif()
 
+  # Add target alias to facilitate unambiguous linking
+  add_library(ESMF::ESMF ALIAS ESMF)
```
danrosen25, line 113: fails if find_package(ESMF) is called multiple times. This is likely in the layered build schemes used to build coupled applications.

Recommended change
```
    if(NOT TARGET ESMF::ESMF)
      add_library(ESMF::ESMF ALIAS ESMF)
    endif()
```

danrosen25 (changes requested):
We'd like for you to cherry-pick this commit onto patch/8.6.1 so that we can bring this fix into the 8.6.1 release. As noted in a comment there is a change requested to check for ESMF::ESMF before adding the alias. Also there needs to be a change to the code and two more files.

- src/addon/ESMX/Comps/ESMX_Data/cmake/FindESMF.cmake
- src/addon/ESMX/Driver/cmake/FindESMF.cmake

AlexanderRichert-NOAA:
@danrosen25 sounds good. I've added the `if` and updated the other two files.

AlexanderRichert-NOAA:
I agree squash-n-merge into develop then cherry pick onto patch branch seems simpler, but I'm happy to do either, just let me know what you guys prefer.

danrosen25:
Got it. I'll finish up testing and put it into both branches.

danrosen25 (approved):

danrosen25:
Done!

Issue #225, opened 2024-02-12
Change needed in FindESMF.cmake


mathomp4:
In working with @AlexanderRichert-NOAA on trying to get the `ESMF::ESMF` target into MAPL, I think I ran into a CMake limitation. Namely, when we build MAPL, CMake autogenerates a `MAPL-targets.cmake` file and in there we see:
```cmake
set_target_properties(MAPL PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include;${_IMPORT_PREFIX}/include/MAPL"
  INTERFACE_LINK_LIBRARIES "MAPL.base;MAPL.generic;MAPL.pfio;MAPL_cfio_r4;MAPL.gridcomps;MAPL.orbit;MAPL.griddedio;MAPL.field_utils;ESMF;NetCDF::NetCDF_Fortran;MPI::MPI_Fortran;\$<\$<BOOL:OFF>:FLAP::FLAP>"
)
```
The issue is that in this we see that `ESMF` not `ESMF::ESMF` is in the `INTERFACE_LINK_LIBRARIES`. As far as I know, the only target that ever gets placed in here is the "actual" library target defined in `FindESMF.cmake`, which is `ESMF`. [Alias targets](https://cmake.org/cmake/help/latest/manual/cmake-buildsystem.7.html#alias-targets) are convenient but not "real".

So, I think what needs to be done is rather `ESMF::ESMF` is the main target supplied by `FindESMF.cmake` and instead we do:
```cmake
  if(NOT TARGET ESMF)
    add_library(ESMF ALIAS ESMF::ESMF)
  endif()
```
essentially "reversing" the current one.

That said, I think I've broken the Baselibs build testing this, so I'll need to debug a bit more...

---

Related PRs:

- https://github.com/GEOS-ESM/ESMA_cmake/pull/354
- https://github.com/GEOS-ESM/MAPL/pull/2599

oehmke:
Hi @mathomp4 , We talked about this at our ESMF meeting today. Would you mind explaining a bit more what this (and associated PRs) is fixing?  Also, is this tested and finished? It wasn't clear to me from the comments if it's done yet. Thanks!

mathomp4:
@oehmke As the main comment says, the issue is that our UFS friends are often picking up an `ESMF` target via our `MAPL-targets.cmake` file that is generated when you build MAPL.

[...]

PR #226, opened 2024-02-13
Update FindESMF.cmake to make ESMF::ESMF main target

mathomp4:
As detailed in #225, our UFS collaborators would like MAPL to provide a `MAPL-targets.cmake` file that uses `ESMF::ESMF` as the target within. From my testing, this can only seem to work if `ESMF::ESMF` is the main target provided by `FindESMF.cmake` and not an `ALIAS`. 

So this PR does that, it essentially makes `ESMF::ESMF` the target the "main" `add_library()` call is run on and then makes `ESMF` the alias.

As far as I can tell, this works. All other needed changes are in GEOS and MAPL.

### DIFF
--- a/cmake/FindESMF.cmake
+++ b/cmake/FindESMF.cmake
@@ -109,7 +109,7 @@ if(EXISTS ${ESMFMKFILE})
     endif()
   endif()
 
-  # Add target alias to facilitate unambiguous linking
+  # Add ESMF as an alias to ESMF::ESMF for backward compatibility
   if(NOT TARGET ESMF)
     add_library(ESMF ALIAS ESMF::ESMF)
   endif()
--- a/cmake/FindESMF.cmake
+++ b/cmake/FindESMF.cmake
@@ -96,22 +96,22 @@ if(EXISTS ${ESMFMKFILE})
       message(WARNING "Static ESMF library (libesmf.a) not found in \
                        ${ESMF_LIBSDIR}. Try setting USE_ESMF_STATIC_LIBS=OFF")
     endif()
-    if(NOT TARGET ESMF)
-      add_library(ESMF STATIC IMPORTED)
+    if(NOT TARGET ESMF::ESMF)
+      add_library(ESMF::ESMF STATIC IMPORTED)
     endif()
   else()
     find_library(ESMF_LIBRARY_LOCATION NAMES esmf PATHS ${ESMF_LIBSDIR} NO_DEFAULT_PATH)
     if(ESMF_LIBRARY_LOCATION MATCHES "ESMF_LIBRARY_LOCATION-NOTFOUND")
       message(WARNING "ESMF library not found in ${ESMF_LIBSDIR}.")
     endif()
-    if(NOT TARGET ESMF)
-      add_library(ESMF UNKNOWN IMPORTED)
+    if(NOT TARGET ESMF::ESMF)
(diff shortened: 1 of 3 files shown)


danrosen25:
Change requested broke the UFS build
```
ld: cannot find -lesmf
make[5]: *** [sorc/sfc_climo_gen.fd/CMakeFiles/sfc_climo_gen.dir/build.make:101: sorc/sfc_climo_gen.fd/sfc_climo_gen] Error 1
make[4]: *** [CMakeFiles/Makefile2:1401: sorc/sfc_climo_gen.fd/CMakeFiles/sfc_climo_gen.dir/all] Error 2
make[4]: *** Waiting for unfinished jobs....
```

mathomp4:
> Change requested broke the UFS build
> 
> ```
> ld: cannot find -lesmf
> make[5]: *** [sorc/sfc_climo_gen.fd/CMakeFiles/sfc_climo_gen.dir/build.make:101: sorc/sfc_climo_gen.fd/sfc_climo_gen] Error 1
> make[4]: *** [CMakeFiles/Makefile2:1401: sorc/sfc_climo_gen.fd/CMakeFiles/sfc_climo_gen.dir/all] Error 2
> make[4]: *** Waiting for unfinished jobs....
> ```

Well, huh. We might need to ask @AlexanderRichert-NOAA for some advice and help...

danrosen25:
I'm testing some more scenarios. Not every FindESMF.cmake file within the ufs-weather-model was consistent. It's possible that I clobbered a FindESMF.cmake file that was customized for a component. I would like to figure this out before making changes to the default library name of ESMF.

mathomp4:
> I'm testing some more scenarios. Not every FindESMF.cmake file within the ufs-weather-model was consistent. It's possible that I clobbered a FindESMF.cmake file that was customized for a component. I would like to figure this out before making changes to the default library name of ESMF.

In GEOS one thing I've had to do (locally) is add aliases for `esmf` as well. But that was due to a long-ago choice local to us. Maybe UFS will need to move to `ESMF::ESMF` internally with this change? I am not sure.

I know from spack testing that if you build with spack, it doesn't seem to matter how many other `FindESMF.cmake` files I have around in GEOS, the one from the ESMF install wins!

danrosen25:
The issue is related to changing one of these three modules, which are already outdated.
- https://github.com/NOAA-EMC/AQM-utils cmake/FindESMF.cmake
- https://github.com/ufs-community/UFS_UTILS cmake/FindESMF.cmake
- https://github.com/noaa-oar-arl/NEXUS cmake/FindESMF.cmake
My guess is that these modules previously defaulted to the static library and the newer FindESMF module will find any ESMF library unless USE_ESMF_STATIC_LIBS is set.

I was able to build the ufs-weather-model with this newer version of the FindESMF module.

danrosen25 (changes requested):
Please copy these changes to the following files:
- src/addon/ESMX/Comps/ESMX_Data/cmake/FindESMF.cmake
- src/addon/ESMX/Driver/cmake/FindESMF.cmake

mathomp4:
> Please copy these changes to the following files:
> 
> * src/addon/ESMX/Comps/ESMX_Data/cmake/FindESMF.cmake
> * src/addon/ESMX/Driver/cmake/FindESMF.cmake

This should be done.

danrosen25 (approved):

PR #229, opened 2024-02-26
Update Schmidt Transform for higher resolution

### LINKED ISSUE #228: Need improved precision for Schmidt transform at high resolution
Bill Putman recently ran into an issue where DP gave bad results with the stretch provided by Schmidt transform.

The improvement is to use higher precision (where available) and use a different value of "TINY".

Quick summary here, but hopefully can point you to an FV PR soon.

```
ifdef NO_QUAD_PRECISION
! 64-bit precision (kind=8)                                                                             
 integer, parameter:: f_p = selected_real_kind(15)
#else
! Higher precision (kind=16) for grid geometrical factors:                                              
 integer, parameter:: f_p = selected_real_kind(20)
#endif
```

```
    near_zero = tiny(near_zero)
```

mathomp4:
Closes #228 

This PR contains code from @wmputman updating the Schmidt Transform code. 

From my tests it does not affect anything if you are not running a stretched grid, but if you are it seems to change the answer (as expected as I'm guessing coordinates are different). I'm guessing your own tests would show this too.

NOTE 1: Let me know if you'd like me to change any names. `f_p` comes from our code and it's not too "ESMF-y". 

NOTE 2: I'm also using:

```fortran
integer, parameter:: f_p = maxval(selected_real_kind(15),selected_real_kind(20))
```

because compilers like NVHPC do not support 128-bit reals (aka quad precision). @tclune was kind enough to tell me I could do this rather than creating a new `#ifdef` and needing to fiddle with the build system (since by Standard, -1 is returned for non-supported kinds). But this does sort of mean things would be "different" on a per-compiler basis without much user notification.

### DIFF
--- a/src/Infrastructure/Util/src/ESMF_UtilCubedSphere.F90
+++ b/src/Infrastructure/Util/src/ESMF_UtilCubedSphere.F90
@@ -35,7 +35,7 @@ module ESMF_UtilCubedSphereMod
   real(ESMF_KIND_R8) , parameter:: torad = pi/180.0          ! convert to radians
   real(ESMF_KIND_R8) , parameter:: missing = 1.e25
   real(ESMF_KIND_R8),  parameter :: JAPAN_SHIFT = pi/18
-  integer, parameter:: f_p = selected_real_kind(15)
+  integer, parameter:: f_p = maxval(selected_real_kind(15),selected_real_kind(20))
   real(ESMF_KIND_R8)    :: stretch               ! Optional stretching factor for the grid 
   logical :: dxdy_area = .false.   ! define area using dx*dy else spherical excess formula
   logical :: latlon = .false.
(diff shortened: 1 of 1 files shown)


oehmke:
We talked about this and decided that we should allow the user to turn this on. Otherwise, the code will give different results depending on whether quad. precision is available in the compiler. The idea is to add another option under coordCalcFlag to allow the user to request this. If it's not available, then an error will be returned.

oehmke:
I talked to NASA team and they said to not worry about putting this in. They have changed how they do things and are now getting coordinates directly from the FV3 model and so don't need to worry about ESMF calculating them. I'll leave this awhile to see if that changes.

anntsay:
Since NASA team already has a workaround.  we need to consider if we should close this PR, or take this PR in at the risk of impacting performance.   another option is to create a branch to add in this PR.. that way, you can also search and pull out this PR and comments and codes.    Prefer to figure out an end point for this PR during 9.0.0. @oehmke to refresh.

theurich:
For safe keeping, `GMAO-SI-Team:228-updated-schmidt`  is locally https://github.com/esmf-org/esmf/tree/gmao-si-team/228-updated-schmidt

oehmke:
Thanks!

mathomp4:
> For safe keeping, `GMAO-SI-Team:228-updated-schmidt` is locally [`gmao-si-team`/228-updated-schmidt](https://github.com/esmf-org/esmf/tree/gmao-si-team/228-updated-schmidt?rgh-link-date=2026-03-04T18%3A19%3A55Z)

@theurich Thanks. I'll ping @tclune as maybe this is just dead code we don't need anymore...

tclune:
Yes - we don't need (want?) ESMF to fix this.  But @mathomp4 we must remain aware.  We're not yet having DYN set the coordinates, and need to do that.  (Basically a Purnendu issue.)

theurich:
Right, we were talking about this during the core call yesterday, and @oehmke did say you were not longer interested in this (plus it could potentially have side effects). So, we do want to close this PR, but also wanted to have our own local copy of the branch, for "just in case", and you guys might decide to delete your fork or branch.

oehmke:
Sounds good. However, we’ll hang on to the branch just in case.

theurich:
Closing now since everyone's in favor of it, and we do have the branch under esmf-org (https://github.com/esmf-org/esmf/tree/gmao-si-team/228-updated-schmidt), for  just-in-case.

oehmke:
Thanks!

PR #231, opened 2024-03-11
Fix FieldGet() index space info arguments to work for all staggers

theurich:
The asynchronous write component under NEPTUNE uses `ESMF_Field` objects on `ESMF_STAGGERLOC_CORNER` for the output options that require regridding. I noticed that the index space info returned by `ESMF_FieldGet()`, e.g. `elementCount`, etc. were incorrect under this situation. Looking at the code I found that it seems it was coded only to work for `ESMF_STAGGERLOC_CENTER` when using Grid, and nodedistrid when using Mesh. This PR attempts to fix this problem, and should support all cases generally by simply using the `DistGrid` of the underlying `Array`.

### DIFF
--- a/src/Infrastructure/Field/src/ESMF_FieldGet.cppF90
+++ b/src/Infrastructure/Field/src/ESMF_FieldGet.cppF90
@@ -374,11 +374,7 @@ type(ESMF_KeywordEnforcer), optional:: keywordEnforcer ! must use keywords below
 !------------------------------------------------------------------------------
     type(ESMF_FieldType), pointer :: ftype
     integer :: localrc, fieldrank
-    type(ESMF_GridDecompType) :: decompType
-    type(ESMF_GeomType_Flag) :: localGeomType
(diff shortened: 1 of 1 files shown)


theurich:
@oehmke - thinking about this fix again, it might actually be something for the 8.6.1 patch? What do you think?

theurich:
Let me change this PR to go into patch/8.6.1, and then I will merge. It will eventually end up in develop via my `feature/neptune-async` branch where I am collecting any fixes needed for one of the NRL projects.

theurich:
Well, I had myself confused... I thought I was basing this branch off of v8.6.0, but instead this is off of develop. Better to just bring it into develop then and cherry-pick into patch/8.6.1. 
It will be going in now...

oehmke:
Sounds good. Let’s remember to merge your neptune branch before 8.7 to get all of that good stuff into the next release… :-)

PR #233, opened 2024-03-13
rename a function inside Zoltan to avoid a name conflict between esmf and shapely

### LINKED ISSUE #419: segfaults due to clash of symbol name "get_geom" between ESMF and Shapely
This issue is to add hopefully sufficient explanation for my pull request #233 from March 2024 (which fixes the issue described here) in order to hopefully give a reason to merge that PR.

----

An external symbol called `get_geom` is provided both by ESMF (inside the Zoltan library which it includes) and also by Shapely.  Because of this clash, if both packages are installed into the same Conda environment, then code that uses Shapely can segfault due to the presence of ESMF in that environment, when an intended call to the `get_geom` inside Shapely ends up executing the one inside ESMF instead.

The fix, which renames `get_geom` to `zoltan_get_geom` inside ESMF, will permit ESMF to co-exist with Shapely in the same conda environment without these segmentation faults happening.  It appears to me that `get_geom` is not a function intended to form part of the ESMF API; it is merely a function used internally by a utility library that ESMF happens to include. It is only called from one place (inside the same source file where it is defined), but is not declared as `static`. So renaming it is unlikely to produce any undesirable side effects.

Just for some context: we provide a Conda environment for data analysis by many users of the [JASMIN](https://jasmin.ac.uk/) facility in the UK, which contains a range of packages including ESMF and Shapely.  While this issue persists, we are having to build a version of ESMF with the same fix applied locally (i.e. renaming that function), and then replace the `libesmf_fullylinked.so` file provided by the Conda package with a symbolic link to the equivalent file in our local build.  This adds significant extra complexity to the build process including a containerized version of the environment, but is currently required as a workaround.  We implemented this following an investigation of segfaults from users, which were also tricky to debug.  In the 14 months since we did this, nobody has reported any issues caused by the fix itself.

----

**Steps to reproduce the problem:**

1) Download and extract the small tarfile [files.tar.gz](https://github.com/user-attachments/files/20431406/files.tar.gz), which contains an environment YAML file, a short Python script and a netCDF data file, all of which are minimal examples.  (Supplied as a tar file because GitHub issues do not allow `.nc` attachments.)  Inspect these files as required.

2) Install and activate a Miniforge base environment if you do not already have one.
```
wget "https://github.com/conda-forge/miniforge/releases/latest/download/Miniforge3-$(uname)-$(uname -m).sh"
bash Miniforge3-$(uname)-$(uname -m).sh
. ~/miniforge3/bin/activate
```

3) Create and activate the environment:
```
mamba env create -n testenv -f testenv.yml
conda activate testenv
```

4) Observe the symbol name clash:
```
$ nm $CONDA_PREFIX/lib/python3.12/site-packages/shapely/lib.cpython-312-x86_64-linux-gnu.so | grep -w get_geom
000000000000d210 T get_geom

$ nm $CONDA_PREFIX/lib/libesmf_fullylinked.so | grep -w get_geom
0000000000347a5e T get_geom
```

5) Run the code (`python test.py`) and observe the segfault.

6) `conda deactivate`, then create a new environment (e.g. use `-n testenv2`), after first commenting out/removing the `xesmf` line from the YAML file.

7) Activate this new environment, and rerun the test code.  Observe that this time a plot is produced.

alaniwi:
Renames the `get_geom` function (not part of the API) which only exists in one source file, in order to avoid a name conflict with a function of the same name in the binary library that forms part of the `shapely` Python package, which can lead to segmentation faults in conda environments where these are both installed.

See the comment included in the changes, which gives more detail.


danrosen25:
Hi @alaniwi,
Can you merge esmf-org/develop into this branch (or rebase, your preference). This will kick off our newly added CI tests. We took a look at the PR today and we are okay with it.

oehmke:
Hi Dan, Yep, I’ll review it. It’s a name change so I think just the CI will be enough. (We’ll also any other issues in the nightlies.)

oehmke (approved):
I slimmed down the comment a bit. Let me know if it looks ok. I think that it looks good now and I'm fine to merge it as is.

oehmke:
Sorry, I wasn’t clear. I was asking you. Sounds good. We’ll merge it in. Thanks for putting in this fix!

PR #234, opened 2024-03-20
Allow setting ESMF_Finalize `endflag` in ESMPy and ESMC/I interfaces

program--:
This PR adds the capability to set the `endflag` parameter in `ESMP_Finalize` to prevent MPI from being finalized at ESMF finalization.

> Support issue for this PR is here: esmf-support#427

The code path followed to support this:
```
ESMP_Finalize
-> ESMC_Finalize
-> ESMCI_Finalize
-> f_esmf_frameworkfinalize_
-> ESMF_Finalize
```

## Additions
- Adds the `ESMC_End_Flag` enumerator to `ESMC_Util.h`, and similarly `esmpy.api.constants.EndAction`.
- Adds `ESMC_FinalizeWithFlag(ESMC_End_Flag endFlag)` to preserve API compatibility with code that calls `ESMC_Finalize`.
- Adds `endFlag` parameter to (with default of `ESMC_END_NORMAL`):
  + `emspy.Manager` constructor
  + `ESMC_FinalizeWithFlag`
  + `ESMCI_Finalize`
  + `f_esmf_frameworkfinalize_` (uses type `type(ESMF_End_Flag)`)

## TODO
The only todo is to validate that unit tests pass, and that this parameter works as intended.


src/Superstructure/ESMFMod/interface/ESMF_Init_C.F90
```diff
@@ -59,14 +59,26 @@ subroutine f_esmf_frameworkinitialize(lang, configFileName, &
...
+       type(ESMF_End_Flag) :: endflag
 
-       call ESMF_Finalize(rc=rc)
+       if (present(keepMpi)) then
+         if (keepMpi) then
+            endflag = ESMF_END_KEEPMPI
+         else
+            endflag = ESMF_END_NORMAL
+         endif
+       endif
```
billsacks, line 80: It looks like this needs an `else` block for the case when `keepMpi` is absent: `endflag` should be set to `ESMF_END_NORMAL` in that case.

src/addon/esmpy/src/esmpy/api/esmpymanager.py
```diff
@@ -121,6 +124,9 @@ def __init__(self, debug=False):
             ESMP_Initialize(logkind=logkind)
             import atexit; atexit.register(self.__del__)
             self.__esmp_initialized = True
+            self.__esmp_keep_mpi = keepMpi
+            if self.__esmp_keep_mpi is None:
+                self.__esmp_keep_mpi = True
```
billsacks, line 129: I think `__esmp_keep_mpi` should be set to False in this case (to be consistent with the documentation, and because that seems like the right thing to do if the caller hasn't set this explicitly to False).

But actually, since the keepMpi argument already defaults to False, I wonder if this None handling is really needed. It looks like maybe you followed the example of the `debug` argument in this respect, but I'm having trouble envisioning how this would get set to `None`. So, unless you have a sense that we need this `None` handling, I'd suggest:
- Getting rid of the handling of `None` here
- Deleting the documentation that `None` means False (line 92)
- Doing a quick test of creating an esmpymanager without setting this explicitly; verifying that it ends up as False as expected

billsacks (changes requested):
Thank you very much for working on this and submitting a PR! I appreciate the time you took to dig down through the layers and make the necessary changes at each layer. I also appreciate your attention to detail in this PR, such as adding documentation in relevant places.

I have two inline comments with things that look like they should be changed a bit. My focus was on the Python pieces, with only a cursory look at the other layers. I'll ask @oehmke to look at the other layers since he's more familiar with that.

billsacks:
> The only todo is to validate that unit tests pass, and that this parameter works as intended.

Please let us know when you have validated that this parameter works as intended - especially that it indeed solves your issue. Thanks again!

src/addon/esmpy/src/esmpy/api/esmpymanager.py
program--: Ah, yes, I think I set this to true by accident when I had a moment of confusing myself about what `keepMpi` meant. In the case of the `None` handling, yeah, you're correct that I just followed the `debug` argument semantics without thinking too much into it. But looking over it again, I agree that it's not really necessary.

src/Superstructure/ESMFMod/interface/ESMF_Init_C.F90
```diff
@@ -59,14 +59,26 @@ subroutine f_esmf_frameworkinitialize(lang, configFileName, &
 
    end subroutine f_esmf_frameworkinitialize
 
-   subroutine f_esmf_frameworkfinalize(rc)
+   subroutine f_esmf_frameworkfinalize(rc, keepMpi)
        use ESMF_CompMod
        use ESMF_InitMod
+       use ESMF_UtilTypesMod
+       use, intrinsic :: iso_c_binding
```
program--, line 66: Just realized a mistake I made -- this implementation is F90, so I don't believe that I can use `iso_c_binding` here (if I remember correctly, it's only F03 and up). May need to pass in an integer instead.

src/addon/esmpy/src/esmpy/api/esmpymanager.py
program--: Resolved in [b3f27d0](https://github.com/esmf-org/esmf/pull/234/commits/b3f27d0420993c0845dd70567a41e8714c5ac326).

src/Superstructure/ESMFMod/interface/ESMF_Init_C.F90
billsacks: It's fine to use iso_c_binding. Despite the file extension being F90, we support this and other F03 features. (I see we use iso_c_binding throughout ESMF, though I'll let @oehmke comment on whether there's a different typical/preferred approach.)

src/addon/esmpy/src/esmpy/api/esmpymanager.py
billsacks: Thanks!

src/Superstructure/ESMFMod/interface/ESMF_Init_C.F90
program--: Resolved in [7249651](https://github.com/esmf-org/esmf/pull/234/commits/72496513fa15a0d867dd15d0be73dab5f0ad3e3a). Thanks for catching this!

src/Superstructure/ESMFMod/interface/ESMF_Init_C.F90
```diff
@@ -59,14 +59,26 @@ subroutine f_esmf_frameworkinitialize(lang, configFileName, &
...
 
        implicit none
 
        integer :: rc
+       logical(kind=c_bool), intent(in), optional :: keepMpi
+       type(ESMF_End_Flag) :: endflag
 
-       call ESMF_Finalize(rc=rc)
+       if(present(keepMpi)) then
```
PhilMiller: In the absent `else` case of `present(keepMpi)`, what's the default value of `endflag` that's getting passed down?

src/Superstructure/ESMFMod/include/ESMC_Init.h
```diff
@@ -125,6 +126,29 @@ extern "C" {
...
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+//-----------------------------------------------------------------------------
+//BOP
+// !IROUTINE: ESMC_FinalizeMPI - Finalize the ESMF Framework and specify whether
+//                               MPI should be finalized
+//
+// !INTERFACE:
```
PhilMiller, line 137: If it's possible to thread an enumerated set of options further up the stack rather than a boolean, that might be good too.

src/Superstructure/ESMFMod/interface/ESMF_Init_C.F90
program--: It would've been an uninitialized value if I remember Fortran semantics correctly, but take a look at [7249651](https://github.com/esmf-org/esmf/pull/234/commits/72496513fa15a0d867dd15d0be73dab5f0ad3e3a), which now sets `endflag` to default to `ESMF_END_NORMAL` and changes it if `keepMpi` is present and true.

program--:
@billsacks @oehmke I've refactored this design slightly to instead propagate `ESMC_End_Flag` at the Python/C/C++ layers, instead of a boolean (as suggested by @PhilMiller), so this should preserve the details found at the Fortran layer. Still haven't gotten a chance to fully test this out, but ESMF compiles correctly on my end with these changes, at least.

src/Superstructure/ESMFMod/include/ESMC_Init.h
program--: Added in [6716d2e](https://github.com/esmf-org/esmf/pull/234/commits/6716d2e82e50a51b5e75f43f4a75dd2090fdd8b8).

src/Superstructure/ESMFMod/include/ESMC_Init.h
```diff
@@ -125,6 +126,29 @@ extern "C" {
...
+  int ESMC_FinalizeMPI(ESMC_Logical keepMpi);
```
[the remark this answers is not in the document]
program--: Changed prototype to `ESMC_FinalizeWithFlag(ESMC_End_Flag endFlag)` in [6716d2e](https://github.com/esmf-org/esmf/pull/234/commits/6716d2e82e50a51b5e75f43f4a75dd2090fdd8b8).

src/Infrastructure/Util/include/ESMC_Util.h
```diff
@@ -58,6 +58,11 @@ enum ESMC_Decomp_Flag {ESMC_DECOMP_INVALID=0,
                        ESMC_DECOMP_BALANCED, ESMC_DECOMP_RESTFIRST,
                        ESMC_DECOMP_RESTLAST, ESMC_DECOMP_CYCLIC};
 
+// end flag
+typedef enum ESMC_End_Flag { ESMF_END_NORMAL=1,
+                             ESMF_END_KEEPMPI,
```
oehmke, line 63: Shouldn't these be ESMC_... (e.g. ESMC_END_NORMAL)?

program--: Ah, yeah. I originally placed `ESMC_End_Flag` in `ESMCI_Util.h`, where those enum values are prefixed with `ESMF_*`, but didn't notice the naming difference.

program--: Fixed in [436de54](https://github.com/esmf-org/esmf/pull/234/commits/436de5445b5f9ffede2bec1361c0776f1ac688aa)

src/Superstructure/ESMFMod/interface/ESMF_Init_C.F90
oehmke: We're slowly moving to iso_c_binding as things are changed or new code is added, so I agree that using it is a good approach.

src/Superstructure/ESMFMod/include/ESMCI_Init.h
```diff
@@ -63,7 +66,7 @@ extern "C" {
                                         ESMC_LogKind_Flag *defaultLogType,
                                         int *rc, ESMCI_FortranStrLenArg count1,
                                         ESMCI_FortranStrLenArg count2);
-   void FTN_X(f_esmf_frameworkfinalize)(int *rc);
+   void FTN_X(f_esmf_frameworkfinalize)(int *rc, ESMC_End_Flag *keepMpi);
```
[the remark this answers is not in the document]
program--: Fixed in [a2d30de](https://github.com/esmf-org/esmf/pull/234/commits/a2d30defaa6880e9d6c2c0a04e71b1bb09019cb4) and made it consistent to the implementation.

src/Superstructure/ESMFMod/interface/ESMC_Init.C
```diff
@@ -143,7 +169,7 @@ extern "C" {
     int localrc;
     
     // todo: it may be better to go directly into F90 instead of using C++
-    localrc = ESMCI_Finalize();
+    localrc = ESMC_Finalize();
```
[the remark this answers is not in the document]
program--: Nope, you're right. I meant to put that back to `ESMCI_Finalize`... I probably need more coffee haha 😅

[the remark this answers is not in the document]
oehmke: :-)

src/Superstructure/ESMFMod/interface/ESMF_Init_C.F90
```diff
@@ -59,14 +59,16 @@ subroutine f_esmf_frameworkinitialize(lang, configFileName, &
...
        use ESMF_InitMod
+       use ESMF_UtilTypesMod
 
        implicit none
 
        integer :: rc
+       type(ESMF_End_Flag) :: endFlag
 
-       call ESMF_Finalize(rc=rc)
+       call ESMF_Finalize(endflag=endFlag, rc=rc)
```
PhilMiller, line 72:
```suggestion
       call ESMF_Finalize(rc=rc, endflag=endFlag)
```
Tiny nitpick, just to keep the arguments in the same order.

src/addon/esmpy/src/esmpy/interface/cbindings.py
```diff
@@ -218,16 +218,22 @@ def ESMP_Initialize(logkind = constants.LogKind.MULTI):
...
+                    released, and, if 'keepMpi' is False, all MPI states
+                    have been cleaned up. This method can only be called
+                    once per execution, and must be preceded by one and
+                    only one call to ESMP_Initialize(). \n
+    Arguments:\n
+        EndAction (optional) :: keepMpi\n
+            Argument Values:\n
+                (default) NORMAL\n
+                KEEP_MPI\n
+                ABORT\n
```
PhilMiller, line 234: Comment text needs to match changed names/type/values

src/Superstructure/ESMFMod/interface/ESMF_Init_C.F90
program--: I think you might've missed the commit message just before 😅 -- putting `rc` at the end is an ESMF style convention

src/addon/esmpy/src/esmpy/interface/cbindings.py
program--: Thanks for catching that, fixed in [1582a32](https://github.com/esmf-org/esmf/pull/234/commits/1582a3262d75f9889b3292206047aa6cce2652d0)

src/Superstructure/ESMFMod/interface/ESMC_Init.C
oehmke, line 1: Hey @billsacks check my reasoning on this...  I think that usually we would want to not have a separate interface for ESMC_FinalizeGetFlag() and just add the flag to the ESMC_Finalize() interface. However, changing that will cause us to have to spread that change everywhere including maybe user code. Also, the ESMC_Finalize() interface doesn't match the ESMC_Initialize() interface in some ways (e.g. no *rc and it doesn't have the multiple argument thing set up). Instead of trying the change all of that now, what I'm thinking is that we keep the ESMC_FinalizeWithFlag() for now and make the doc non-public, so users won't see it. Then at some point in the future  (maybe before 8.7?) we make all the changes at once to have ESMC_Finalize() match ESMC_Initialize and bring in the endFlag?  I'd rather change a public interface all at once, and I don't want to hold this up while we figure out that...

src/Superstructure/ESMFMod/interface/ESMC_Init.C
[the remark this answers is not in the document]
program--: Fixed in [bdc194a](https://github.com/esmf-org/esmf/pull/234/commits/bdc194a3e97979773867ff38a16fede526cabeaa). Thanks!

src/addon/esmpy/src/esmpy/interface/cbindings.py
```diff
@@ -218,16 +218,22 @@ def ESMP_Initialize(logkind = constants.LogKind.MULTI):
...
-def ESMP_Finalize():
+def ESMP_Finalize(endFlag = constants.EndAction.NORMAL):
     """
     Preconditions: ESMF has been initialized.
     Postconditions: ESMF has been finalized, all heap memory has been
-                    released, and all MPI states have been cleaned up.
-                    This method can only be called once per execution,
-                    and must be preceded by one and only one call to
-                    ESMP_Initialize(). \n
+                    released, and, if 'keepMpi' is False, all MPI states
```
oehmke: Needs to be changed to say something about endFlag equaling Normal.

oehmke (commented):
Hi Justin, 

  First, I just want to echo what Bill said, thanks for all of your careful work on this!  I think that overall this looks great except for the few small changes that I mentioned. We are still talking a little about the one interface, but I think that other than that it looks good to 
me. Let us know how it goes with your testing. 

Thanks,
- Bob

program--:
@oehmke Awesome, glad to hear! I'll finish up those small changes today. I've also tested with a small toy program so far that ensures the C API works as expected (not exactly a unit test, but close?)

<details><summary>C Example Program</summary>

```c
#include <stdio.h>
#include <assert.h>
#include <string.h>

int state_ = 0;

extern int PMPI_Finalize();

// intercept MPI calls to output
int MPI_Finalize() {
    printf("\nMPI_Finalize called\n");
    PMPI_Finalize();
    state_ = 1;
    return 0;
}

int MPI_Abort(int comm, int errorcode) {
    printf("\nMPI_Abort called\n");
    state_ = 10;
    return 0;
}

#include <mpi.h>
#include "ESMC.h"

int main(int argc, const char* argv[]) {

    if (argc < 2) {
        printf("Expected one of: keep_mpi, abort, normal\n");
        return 0;
    }

    MPI_Init(NULL, NULL);

    ESMC_End_Flag end = ESMC_END_NORMAL;

    if (strcmp(argv[1], "keep_mpi") == 0) { 
        end = ESMC_END_KEEPMPI;
    } else if (strcmp(argv[1], "abort") == 0) {
        end = ESMC_END_ABORT;
    } // doesn't actually check for normal

    int rc;
    ESMC_Initialize(&rc, ESMC_ArgLast);
    ESMC_FinalizeWithFlag(end);

    switch(end) {
        case ESMC_END_KEEPMPI:
            printf("Called with ESMC_END_KEEPMPI\n");
            assert(state_ == 0);
            MPI_Finalize();
            break;
        case ESMC_END_NORMAL:
            printf("Called with ESMC_END_NORMAL\n");
            assert(state_ == 1);
            break;
        case ESMC_END_ABORT:
            printf("Called with ESMC_END_ABORT\n");
            assert(state_ == 10);
            break;
    }

    return 0;
}

```

In this case, I built with clang, gfortran, and (intel) mpich.

Compiling this to `test_esmf_endflag` provides the following outputs:
```sh
# normal
./test_esmf_endflag normal
#> MPI_Finalize called
#> Called with ESMC_END_NORMAL

# keep_mpi
./test_esmf_endflag keep_mpi
#> Called with ESMC_END_KEEPMPI
#> MPI_Finalized called

# abort
./test_esmf_endflag abort
#> MPI_Abort called
#> Called with ESMC_END_ABORT
```

</details>

I will verify with Python as well, but I expect I'll see the same results 🙂 Thanks again for the review and suggestions!

src/addon/esmpy/src/esmpy/interface/cbindings.py
program--: I think this was in reference to an older commit, but I revised the docstring again to be explicit about MPI cleanup only when endFlag is set to `NORMAL` (and not necessarily when it's set to something other than `KEEP_MPI`). Fixed in [0f7b9b4](https://github.com/esmf-org/esmf/pull/234/commits/0f7b9b4a215f8648c472a48d3a262ada19b09bab).

src/Superstructure/ESMFMod/interface/ESMC_Init.C
billsacks: > I think that usually we would want to not have a separate interface for ESMC_FinalizeGetFlag() and just add the flag to the ESMC_Finalize() interface. However, changing that will cause us to have to spread that change everywhere including maybe user code. Also, the ESMC_Finalize() interface doesn't match the ESMC_Initialize() interface in some ways (e.g. no *rc and it doesn't have the multiple argument thing set up). Instead of trying the change all of that now, what I'm thinking is that we keep the ESMC_FinalizeWithFlag() for now and make the doc non-public, so users won't see it. Then at some point in the future (maybe before 8.7?) we make all the changes at once to have ESMC_Finalize() match ESMC_Initialize and bring in the endFlag? I'd rather change a public interface all at once, and I don't want to hold this up while we figure out that...

@oehmke - yes, that makes sense to me, if this is potentially breaking backwards compatibility. I haven't yet come to understand how ESMF handles optional arguments in the C interface, but looking at ESMC_Initialize, I see that there is a required ESMC_ArgLast argument (even if calling it without any optional arguments), so I see how following that approach would be backwards-incompatible. I had hoped that there could be a way to add an optional argument without breaking backwards compatibility, but maybe that isn't possible for the C interface.

@PhilMiller 's suggestion of putting this in 9.0 also makes sense to me... we can add this to the list of API-breaking things we're considering for 9.0. Does that make sense to you @oehmke or do you see a way this could be done such that it would be unlikely to interfere with (much) user code and so could be done before 9.0?

I'm guessing that the reason for making ESMC_FinalizeWithFlag non-public is because, if it's public, then removing it would be yet another API-breaking change. If so, I think I like the idea of keeping it non-public.

src/Superstructure/ESMFMod/include/ESMCI_Init.h
```diff
@@ -51,7 +52,9 @@ int ESMCI_Initialize(ESMC_CalKind_Flag defaultCalendar=ESMC_CALKIND_NOCALENDAR);
 int ESMCI_Initialize(int argc, char **argv,
   ESMC_CalKind_Flag defaultCalendar=ESMC_CALKIND_NOCALENDAR);
 
-int ESMCI_Finalize(void);
+int ESMCI_Finalize();
```
billsacks, line 55: Was this change from `ESMCI_Finalize(void)` to `ESMCI_Finalize()` intentional? I see that `ESMCI_Init.C` still defines it as `ESMCI_Finalize(void)`; should they be kept consistent?

billsacks (changes requested):
This is looking great. I noticed one change where I wonder if the difference is intentional or an accident of some back-and-forth changes. Other than that, this looks good to me, but I'll also wait for a final review from @oehmke , since he's much more familiar with what's needed at the various layers than I am.

src/Superstructure/ESMFMod/include/ESMCI_Init.h
program--: Purely an accident! I'll get that reverted, thanks for spotting that.

src/Superstructure/ESMFMod/interface/ESMC_Init.C
oehmke: @billsacks - Yeah, I was thinking about the backward compatibility. As you guessed, that's also why I was thinking of making the ESMC_FinalizeWithFlag() non-public.  That way, if a user needs it, we have the option of pointing them to it, but we're still limiting official access if we want to merge the interfaces in the future. I think that @PhilMiller 's suggestion about 9.0 makes sense. If we start having a lot of users asking for the endFlag capability through the C-API then we can consider making it public, but as far as we know we don't have a lot of people using that, so I think that it makes sense to just defer it to 9.0.

billsacks: > Yeah, I was thinking about the backward compatibility. As you guessed, that's also why I was thinking of making the ESMC_FinalizeWithFlag() non-public. That way, if a user needs it, we have the option of pointing them to it, but we're still limiting official access if we want to merge the interfaces in the future. I think that @PhilMiller 's suggestion about 9.0 makes sense. If we start having a lot of users asking for the endFlag capability through the C-API then we can consider making it public, but as far as we know we don't have a lot of people using that, so I think that it makes sense to just defer it to 9.0.

Thanks - that all makes sense to me.

PhilMiller: That all sounds good to me. It'll work for our use case, and we can adapt to a future API change in 9.0 if needed.

src/addon/esmpy/src/esmpy/interface/cbindings.py
oehmke: Looks good! Thanks!

billsacks (approved):
This all looks good to me now - thank you! @program-- I wanted to also thank you for showing the test program you used for this. And @PhilMiller thank you for your contributions to this review and discussion as well!

I ran the full ESMF and ESMPy testing on my Mac, and this looks good.

I'll wait on final approval from @oehmke before merging.

PhilMiller:
Thank you both for working with us on this. It will help us a lot in integrating ESMF functionality in the NWM NextGen system.

Can I ask when you think this will make it into a release, once it's merged?

oehmke:
Thanks to both of you for your careful work on this!  

We are planning on having our next ESMF release (8.7) in early summer and this should be included in that. However, once this is merged and has gone through our nightly tests, then I can make you a beta snapshot tag to use, if that would be helpful.

PhilMiller:
Ok, that sounds good, thank you. I think the beta tag would potentially be useful.

oehmke (approved):
I took a final look at this and everything looks good to me. I'll probably merge this tomorrow(3/26) unless anyone has any final objections. Thanks again!

jduckerOWP:
@oehmke can you confirm to the community here the earliest version of ESMF that this PR has been merged to? Thank you all very much for your efforts on this front, it's been a great help to the NOAA-OWP community.

oehmke:
It’s great to hear that it’s been helpful! The earliest tag that contains that PR is v8.7.0b06. It’ll also be out in ESMF 8.7 which is expected late summer.

PR #243, opened 2024-05-08
Fixes #242 Update ESMF_VM.F90

### LINKED ISSUE #242: BUG:  segfault with ssiMap argument to ESMF_VMGet() is broken
The following snippet of code contains an erroneous reference to an optional argument `petCount`.
https://github.com/esmf-org/esmf/blob/e29a9710c480f8047c896b6c29de5c3db608b89e/src/Infrastructure/VM/interface/ESMF_VM.F90#L5356-L5363

The reference should instead be to `petCountArg`

As a workaround I can include the unused optional argument until this is fixed.

I'll submit a PR shortly.

tclune:
Trivial fix fortunately.


theurich (approved):
Good find!

PR #249, opened 2024-05-17
Fix spack build action

uturuncoglu:
This PR aims to fix spack build action and includes following changes. 

- The new repo for apt get is added to allow Ubuntu to find gfortran-13 package correctly.
- The g++-13 package is also added to the list of required packages explicitly since Ubuntu was not installing it by default and spack package manager had issue to find the g++13 compiler.

The chnages in action is tested under my own personal fork. The successful run can be seen in https://github.com/uturuncoglu/esmf/actions/runs/9123456143/job/25085911284.

### DIFF
--- a/.github/workflows/test-build-spack.yml
+++ b/.github/workflows/test-build-spack.yml
@@ -83,7 +83,7 @@ jobs:
         sudo add-apt-repository ppa:ubuntu-toolchain-r/test
         sudo apt-get -qq install tar unzip file curl gringo
         sudo apt-get -qq install build-essential binutils-dev gfortran gdb
-        sudo apt-get -qq install gfortran-12 gfortran-13
+        sudo apt-get -qq install gfortran-12 g++-13 gfortran-13
         sudo apt-get -qq install python3-dev
 
     # restore Intel oneAPI compiler installation from cache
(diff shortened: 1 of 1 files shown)


.github/workflows/test-build-spack.yml
```diff
@@ -80,9 +80,10 @@ jobs:
     - name: Install Core Development Tools
       run: |
         sudo apt-get -qq update
+        sudo add-apt-repository ppa:ubuntu-toolchain-r/test
```
danrosen25, line 83: This is an untrusted PPA (Personal Package Archive). This introduces a security risk if any sensitive information (keys, passwords, etc) are ever entered into the GitHub runner ... including the ssh keys used by git clone and push actions. Can you find an alternative or skip using gfortran-13?

danrosen25 (changes requested):
The untrusted ppa is a security risk for any sensitive information. A scammer could do anything to the code obtained from the untrusted ppa including intercept private keys.

uturuncoglu:
@danrosen25 Let me check. Maybe I could find another solution. We could drop 13 support but it is good to check recent version of gnu compiler.

uturuncoglu:
@danrosen25 Thanks for your suggestion. I changed the runner to Ubuntu 24.04. It has 

gcc@=12.3.0
gcc@=13.2.0
gcc@=14.0.1

So, I think we could go with it. The action uses most recent one by default. I am testing now and if it works I'll change it in the PR.

danrosen25 (approved):

PR #267, opened 2024-07-23
ESMPy: Fix type cast to support large Arrays

billsacks:
In a few places, the size of an Array was being cast to a c_int. This caused an overflow if the size exceeded 32 bits. An example of the issue is reported here, and I have reproduced this:
https://github.com/orgs/esmf-org/discussions/263

From some reading, I *think* the correct cast here should be to ssize_t, which resolves to a long int (64 bits) on a 64-bit architecture. I am not 100% positive of this, but this seems consistent with what I have found; e.g., see:
- Argument list in the PyMemoryView_FromMemory function: https://docs.python.org/3/c-api/memoryview.html
- https://docs.python.org/3/c-api/intro.html#c.Py_ssize_t
- https://stackoverflow.com/questions/20987390/cython-why-when-is-it-preferable-to-use-py-ssize-t-for-indexing

### DIFF
--- a/src/addon/esmpy/src/esmpy/util/esmpyarray.py
+++ b/src/addon/esmpy/src/esmpy/util/esmpyarray.py
@@ -29,7 +29,7 @@ def ndarray_from_esmf(data, dtype, shape):
     if sys.version_info[0] >= 3:
         buffer = ct.pythonapi.PyMemoryView_FromMemory
         buffer.restype = ct.py_object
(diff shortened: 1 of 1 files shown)


billsacks:
ESMPy testing passes with this change on my Mac.

billsacks:
@rokuingh - I'm going ahead and merging this, but if you want to look at this later and notice any issues, I can change it.

Issue #279, opened 2024-08-26
Install FindESMF.cmake into <ESMF_INSTALL_PREFIX>/cmake


danrosen25:
During the `make install` step provide `FindESMF.cmake` into the installation of ESMF under a `<ESMF_INSTALL_PREFIX>/cmake` directory.

Yes, I realize this is strange because FindESMF.cmake is intended to find the ESMF_INSTALL_PREFIX location given ESMF_ROOT or ESMFMKFILE as hints.

danrosen25:
@theurich - Do you have any thoughts on this before I proceed?

theurich:
It is strange, and previously when this has come up, I always recommended not doing it. 
However, I am starting to realize that in practice people might expect to find `FindESMF.cmake` within the ESMF installation. Those people might not go look for it on GitHub to find it under the ESMF repo (which is actually where they should get it). So I am fine with this going under the installation dir, inside a `cmake` directory now, in order to help out those folks that go look for it there.
Maybe we should also put a `README` file inside of the `cmake` dir under the installation dir, where we explain how the `FindESMF.cmake` is intended to be used. I.e. it is an example of a `FindESMF.cmake`, which a user would use in their own project by copying it to their own project and use it from there. They might need to make modifications on their own version if it isn't exactly what they need it to do, but it should be a good starting point.

danrosen25:
@theurich 
What do you think of this documentation?
https://github.com/danrosen25/esmf/tree/feature/install_cmake/cmake
I'll bring this branch over to esmf-org once edits are complete. Then I can put it on the project board to integrate this into develop for 8.8.0

PR #282, opened 2024-08-29
Install CMake modules into ESMF_INSTALL_CMAKEDIR

danrosen25:
* add how-to use documentation to FindESMF.cmake
* copy *.cmake files into ESMF_INSTALL_CMAKEDIR (default ESMF_INSTALL_PREFIX/cmake)
* update ESMF user's guide
* add README.md in cmake directory (do not copy into installation as this might override README.md fill in a user's folder)

Resolves: https://github.com/esmf-org/esmf/issues/279


billsacks (approved):
Looks great!

src/doc/ESMF_install.tex
```diff
@@ -1322,6 +1329,7 @@ \subsection{Installing the ESMF}
 \item {\tt ESMF\_INSTALL\_MODDIR} -- where to install Fortran module files.
 \item {\tt ESMF\_INSTALL\_BINDIR} -- where to install application files.
 \item {\tt ESMF\_INSTALL\_DOCDIR} -- where to install documentation files.
+\item {\tt ESMF\_INSTALL\_CMAKEDIR} -- where to install cmake module files.
```
theurich, line 1332: None of the other items have a period (.) at the end of sentence. Maybe all should get one? Either way fine with me, just should be consistent.

cmake/README.md
```diff
@@ -0,0 +1,45 @@
+# CMake Modules Provided by ESMF
+
+The ESMF team provides CMake[1] modules that are useful when ESMF is needed
+within a CMake based build system. These modules can find libraries, create
+targets, and set variables within the scope of a CMake build. They may be
+used as is or customized for application specific build systems.
```
theurich, line 6: Could we add a sentence at the end of this paragraph to point out that CMake module files are typically kept with the project that use them? Maybe:
"Typically CMake module files are copied into the project that use them. They become part of the project and are maintained with it."

theurich (changes requested):
Looks very good. Just a couple of small suggestions.

src/doc/ESMF_install.tex
theurich: I could have sworn that there was no period when I reviewed this... but all I can say is that I definitely need (better) glasses! This looks good as is!

theurich (approved):
Looks good to me now. Thanks for adding the extra sentences to the usage discussion.

We should discuss during the core call whether this could go into 8.7.0, or should wait for 8.8.0.

danrosen25:
Bring into 8.7.0 after the ESMX Fix

Issue #281, opened 2024-08-28
Fix ESMX - link_module_paths, two stage appConf


danrosen25:
While attempting to link Julia_LIBRARY into ESMX_SingleModelInJuliaProto a few bugs were discovered

- add module paths wrong variable name `ESMX_CMAKE_MODULE_PATHS` should be `ESMX_LINK_MODULE_PATHS`
- due to variable expansion ${Julia_LIBRARY} cannot be used until after find_package(Julia)
- link_libraries can't find a library if the full path is given

Code changes are ready for review on fork: https://github.com/danrosen25/esmf/tree/fix/esmx_app_link

Julia prototype changes once ESMX changes are merged
[julia_ld_cleanup.patch](https://github.com/user-attachments/files/16787842/julia_ld_cleanup.patch)

danrosen25:
Hi @billsacks I completed the changes needed other than renaming the user facing configuration options. We should discuss these configuration option name changes with @theurich before proceeding

- link_module_paths -> append_module_paths
- link_packages -> find_packages

billsacks:
This is great @danrosen25 - thank you very much for getting this working! I like this way of handling this linking to Julia.

In addition to your changes, we should change this section of the README in nuopc-app-prototypes; I can do that once we're ready to bring in these changes:

https://github.com/esmf-org/nuopc-app-prototypes/tree/develop/ESMX_SingleModelInJuliaProto#specifying-build-options-needed-to-invoke-julia-code

PR #283, opened 2024-08-29
Fix ESMX module paths and library linking using packages

danrosen25:
* fix bug in variable name for ESMX_LINK_MODULE_PATHS
* add capability to find packages before expanding variables, which allows you to pick up variables from find_package calls
* add capability to provide a full path to link library

Needs Attention: Do we want to rename configurations options

- link_module_paths -> append_module_paths
- link_packages -> find_packages

Resolves: https://github.com/esmf-org/esmf/issues/281


billsacks (approved):
Thanks again for your work on this @danrosen25 ! This will definitely be a better way to bring in Julia components!

danrosen25:
Documentation for `link_packages` needs to be fixed. It needs to **explicitly** call out the link_libraries step is needed.

theurich (approved):
Thank you!!

danrosen25:
Documentation fix, run ESMX tests, and bring into develop for `8.7.0`.

PR #285, opened 2024-09-03
New system test that tests ESMC_FinalizeWithFlag() as well as some basic C-API regridding functionality

oehmke:
This PR adds a new system test that tests some basic ESMF C-API functionality by doing a simple regridding. It also tests ESMC_Initialize(), ESMC_Finalize(), and ESMC_FinalizeWithFlag().


theurich (commented):
After looking through the changes I thought it would be better to bring in more clean pure C STest support. I.e. Prefix ESMC, and suffix lower case .c. Those changes would be hard to discuss in this PR, and therefore I opened PR #290 aimed into this branch. Best to review them there.

PR #286, opened 2024-09-05
Add ESMF development tests for PR and branches

danrosen25:
* fix non-exhaustive tests in ESMF_FieldIOUTest.F90
* development tests automatically execute for PR into develop
* development tests can be executed manually for any branch

| system       | compiler          | comm    | exhaustive | bopt | tracing | all_tests |
| ------------ | ----------------- | ------- | ---------- | ---- | ------- | --------- |
| ubuntu@22.04 | gfortran@11       | mpiuni  | ON         | O    | ON      | ~12m      |
| ubuntu@22.04 | gfortran@12       | mpich   | OFF        | O    | ON      | ~22m      |
| macos@14     | gfortran@13       | mpiuni  | OFF        | O    | OFF     | ~10m      |
| macos@14     | clang+gfortran@14 | openmpi | ON         | g    | ON      | ~17m      |

### DIFF
--- a/.github/workflows/development-tests.yml
+++ b/.github/workflows/development-tests.yml
@@ -1,0 +1,250 @@
+name: Development Tests
+
+on:
+  schedule:
+    - cron: "15 8 * * 3"
+  pull_request:
+    branches:
+      - develop
+  workflow_dispatch:
+
+jobs:
+  development-test:
+    name: ${{matrix.config.desc}}
+    runs-on: ${{matrix.config.osys}}
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+        - {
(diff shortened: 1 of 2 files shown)


billsacks (approved):
This is fantastic! Thanks again for putting this in place - this will be great to have! And this seems like a good starting point for the coverage matrix.

oehmke:
I think that for now, let’s keep this as a PR, manually triggered, and weekly (to keep the cache working). After 8.7 and seeing how it goes, we can think about every push. What do you think?

billsacks:
I don't have strong feelings on the frequency of running these. @oehmke 's suggestion seems reasonable for now.

PR #289, opened 2024-09-05
Potentially Fixes #288 ESMF_ArrayCreateGetUTest

### LINKED ISSUE #288: farrayPtr2DCpy (a.k.a. arrayDup) is not initialized in ESMF_ArrayCreateGetUTest.F90
https://github.com/esmf-org/esmf/blob/82ba326579f89b8946c99b2f953d37223b28d7e0/src/Infrastructure/Array/tests/ESMF_ArrayCreateGetUTest.F90#L408

### Test Failure 
Verify Array vs Array Copy (ALLOC) no data copy, [ESMF_ArrayCreateGetUTest.F90, line 411](https://github.com/esmf-org/esmf/blob/82ba326579f89b8946c99b2f953d37223b28d7e0/src/Infrastructure/Array/tests/ESMF_ArrayCreateGetUTest.F90#L398):  Unexpected data copy
`if (abs(farrayPtr2D(i,j)-farrayPtr2DCpy(i,j)) < 1.d-10) dataCorrect=.false.`

Initialization for farrayPtr2DCpy never happens because it is created with ESMF_DATACOPY_ALLOC
`arrayDup = ESMF_ArrayCreate(array, datacopyflag=ESMF_DATACOPY_ALLOC, rc=rc)`
Initialization for farrayPtr2D
`farrayPtr2D = real(localPet+10, ESMF_KIND_R8)`

This causes intermittent failures

danrosen25:
Potentially Fixes #288 ESMF_ArrayCreateGetUTest where data in arrayDup is not initialized and compared to data in array.
* fill array with random number 0 - 1000
* fail if every element in array matches

### DIFF
--- a/src/Infrastructure/Array/tests/ESMF_ArrayCreateGetUTest.F90
+++ b/src/Infrastructure/Array/tests/ESMF_ArrayCreateGetUTest.F90
@@ -369,7 +369,8 @@ program ESMF_ArrayCreateGetUTest
   !NEX_UTest_Multi_Proc_Only
   write(name, *) "ArrayCreate from Copy (ALLOC), 2D ESMF_TYPEKIND_R8 Test"
   write(failMsg, *) "Did not return ESMF_SUCCESS"
-  farrayPtr2D     = real(localPet+10, ESMF_KIND_R8)  ! fill with data to check
+  call random_number(farrayPtr2D) ! fill with data to check
+  farrayPtr2D = farrayPtr2D * 1000.0_ESMF_KIND_R8
   arrayDup = ESMF_ArrayCreate(array, datacopyflag=ESMF_DATACOPY_ALLOC, rc=rc)
   call ESMF_Test((rc.eq.ESMF_SUCCESS), name, failMsg, result, ESMF_SRCLINE)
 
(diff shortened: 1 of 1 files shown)


theurich (commented):
This looks very close to what I was thinking. Thank you for starting this PR.

One concern I have is with `hugh()`. My worry is that the largest representable number might not be so special, i.e. it might be not unlikely to hit it in uninitialized memory. I don't know that it is true, but that is just my feeling. Subtracting localpet might not change anything about it, since `hugh` might be so large that subtracting something O(1) makes no difference.

With that in mind, I was more thinking of initializing maybe in the more reasonable range of say 1000*(localPet+1), and then in addition start change maybe by 0.1 for each local element increment. Hitting all of those matching seems super unlikely to me.

danrosen25:
@theurich 
How about the new changes, where I fill the array with a number between 0 and 1000 (inclusive of 0 and exclusive of 1000)?

theurich:
> @theurich How about the new changes, where I fill the array with a number between 0 and 1000 (inclusive of 0 and exclusive of 1000)?

@danrosen25 I like this version! However, I am going to tag @billsacks here, since he had valid concerns about use of "randomness" in testing before. I agree with him that it is generally a bad idea to have randomness in tests, because you loose reproducibility, which is a pain to deal with when things are failing. However, here I think the situation is random already, due to the uninitialized data.... so in my mind I do not see a problem with use of `random_number` in filling the array that the other random data is compared against. Still, I am interested to see what Bill thinks about this situation.

billsacks (changes requested):
Thanks for tagging me on this, and sorry for my delay in responding.

I agree with all points in @theurich 's last comment - that we should generally try to avoid randomness, but also that this situation is special in that the only purpose of these values is to compare with uninitialized data, which is already effectively random. So I support this change. I do feel like a *slightly* more ideal change would be to initialize farrayPtr2D with a hard-coded list of pseudo-random values, but I feel like this use of random_number is okay here. I'd just say that ideally there would be a comment before the random_number call along these lines:

```
! Generally we want to avoid using random_number in unit tests. But here the purpose of the initialization is to compare against uninitialized data, which is already effectively random. So we use random number initialization for convenience and because it doesn't lead to much additional randomness of this test against uninitialized data.
```

danrosen25:
@billsacks 
If you're okay with the comment I added then I'll squash merge this into develop. Thanks!

billsacks (approved):
Looks good - thank you very much for this @danrosen25 !

Issue #237, opened 2024-04-08
Sporadic failures in ESMF_CompTunnelUTest and ESMF_CompTunnelEx on Mac systems


billsacks:
I have noticed sporadic failures in ESMF_CompTunnelUTest and ESMF_CompTunnelEx on my Mac (green). I have seen this when running multiple test suites at once, and have seen this at least once in nightly cron-based testing.

I first wondered if this had something to do with timeouts. @theurich says that these tests do have timeout values, but they're pretty large, so this is unlikely to be a performance-related issue.

@theurich says that what makes these tests unique is that they set up ports / sockets to make connections. If the port is used, then you have a problem – e.g., due to a hanging process. This can also come up if you run multiple instances of ESMF tests at the same time. Or a process from before could have crashed and there could be some hanging stuff. We have seen this periodically on Mac systems.

For now we'll just note that this can sometimes be an issue but close it as a wontfix. If this becomes a frequent issue, we could consider introducing a flag for testing that disables anything that could fail when running simultaneous test suites; these are the only tests that @theurich can think of that have this issue, but there could be others in the future.

billsacks:
As noted above, closing as a wontfix - I just opened this to document the issue.

PR #294, opened 2024-09-10
Last minute NUOPC Connector items for ESMF 8.7

theurich:
Adding connection options support for regridding method "conserve_2nd" and extrapolation method "nearest_d". NRL NEPTUNE developers are hoping to use "conserve_2nd" with ESMF 8.7.

This PR also removes old code from NUOPC_Connector that has been inactive for several releases now. This is code that had been under the `#ifdef USE_ESMF_RHL #else` branches.


danrosen25 (approved):
Looks good as long as the tests pass. We don't have explicit tests for `remapMethod` and they would be located with the NUOPC app prototypes. I didn't know that type RHL was moved into the ESMF library as ESMF_RHL. We should consider moving this to ESMF_RouteHandleList in the future but that's a bigger cleanup.

danrosen25:
@theurich @billsacks 
I looked at the failure and it's an issue I've seen before. There was an [issue](https://github.com/esmf-org/esmf/issues/237) for it but it was closed without resolving.

```
20240910 215612.746 INFO             PET4  FAIL  Check time delay on non-blocking Component D Finalize/wait, ESMF_CompTunnelUTest.F90, line 1283:  Incorrect time delay
20240910 215612.746 INFO             PET4  delayTime (wait) =    2.9923109999999973
```
What's happening is that the time between ESMF_GridCompFinalize and ESMF_GridCompWait is less 3 seconds but the test assumes it will take greater than 3 seconds. Mac OS systems are finishing too quickly!
https://github.com/esmf-org/esmf/blob/9379ba6a7ea05a870c3c320be556aa76b23df470/src/Superstructure/Component/tests/ESMF_CompTunnelUTest.F90#L1255
https://github.com/esmf-org/esmf/blob/9379ba6a7ea05a870c3c320be556aa76b23df470/src/Superstructure/Component/tests/ESMF_CompTunnelUTest.F90#L1267

A possible solution could be to change the test to be something much smaller, like 1 second. I'm not sure why it's 3 seconds.
https://github.com/esmf-org/esmf/blob/9379ba6a7ea05a870c3c320be556aa76b23df470/src/Superstructure/Component/tests/ESMF_CompTunnelUTest.F90#L1283

theurich:
There are several reasons the CompTunnelUTest can fail. https://github.com/esmf-org/esmf/issues/237 discusses the socket issues. Here it is a timing issue - which is different.
There is a call to `ESMF_VMWtimeDelay` with 3 seconds in the test. The 3s are set by macro `SLEEPTIME`. For some reason either the clock precision is not determined correctly or there is some drifting of the clock or the WtimeDelay.. but just sometimes. I guess we could build in a larger tolerance than just 2x the clock precision. But it is a strange thing why Macs would need that.

oehmke (approved):
Looks good to me.

danrosen25:
@theurich 
Okay, I see. I did think it was odd that finalize was expected to take longer than 3 seconds but now I see the call to `ESMF_VMWtimeDelay`. I assume that the `ESMF_VMWtimeDelay` calls sleep under the hood? It doesn't seem possible that the time between ESMF_VMWtime calls is less than 3 seconds. So I assume there's something going on within the system. I believe the runners are VMs so it's possible that the VM server messes around with the system clock?

PR #297, opened 2024-09-12
Implement the ESMX `link_options` application option

theurich:
Sometimes I have the need to specify a link option for when the ESMX executable is linked. E.g. `-check_mpi` to turn on IntelMPI check features. I can hack this into one of the components' cmake module files, but it would be cleaner if it could be specified as an esmx application option in `esmxBuild.yaml`. This PR adds this feature by adding the `link_options` application option.


danrosen25 (approved):
Simple, looks good!

Issue #284, opened 2024-09-03
Timemgr using sprintf with the output string also used in the input


oehmke:
Atanas  ran into an issue with a garbled time string and traced it down to a sprintf call in the time manager. It looks like there are several instances of sprintf() with the output string being also used as an input. It sounds like in general this shouldn't be done, but that it sometimes works (depending on the implementation of sprintf()). We should fix all of these, so we don't run into random issues. An example of one of these is in .../TimeMgr/src/ESMCI_Time.C line 1482.

billsacks:
I'm happy to take this on.

I just did a regex search to try to identify the places that need to be fixed. This would miss anything where a sprintf call is split across multiple lines, but hopefully is at least a good start: The regex is `sprintf *\(([^,]+) *,.*\1`. There are a number of false positives, but it picks up some things that should be fixed:

https://github.com/esmf-org/esmf/blob/82f8052f9cfee57e81a2c7a74e5dfa05870b7b93/src/Infrastructure/Mesh/src/Legacy/ESMCI_MeshMerge.C#L1183

https://github.com/esmf-org/esmf/blob/82f8052f9cfee57e81a2c7a74e5dfa05870b7b93/src/Infrastructure/Mesh/src/Legacy/ESMCI_MeshMerge.C#L1186

https://github.com/esmf-org/esmf/blob/82f8052f9cfee57e81a2c7a74e5dfa05870b7b93/src/Infrastructure/TimeMgr/src/ESMCI_Time.C#L1471

https://github.com/esmf-org/esmf/blob/82f8052f9cfee57e81a2c7a74e5dfa05870b7b93/src/Infrastructure/TimeMgr/src/ESMCI_Time.C#L1473

https://github.com/esmf-org/esmf/blob/82f8052f9cfee57e81a2c7a74e5dfa05870b7b93/src/Infrastructure/TimeMgr/src/ESMCI_Time.C#L1480

https://github.com/esmf-org/esmf/blob/82f8052f9cfee57e81a2c7a74e5dfa05870b7b93/src/Infrastructure/TimeMgr/src/ESMCI_Time.C#L1482

https://github.com/esmf-org/esmf/blob/82f8052f9cfee57e81a2c7a74e5dfa05870b7b93/src/Infrastructure/TimeMgr/src/ESMCI_TimeInterval.C#L3142

[...]

Issue #295, opened 2024-09-11
Incorrect time delay ESMF_CompTunnelUTest - Mac OS w OpenMPI


danrosen25:
ESMF_CompTunnelUTest fails intermittently due to the time delay between ESMF_GridCompFinalize and ESMF_GridCompWait for an ESMF_SYNC_NONBLOCKING call being less than 3 seconds. This doesn't make sense as the finalize subroutine calls ESMF_VMWtimeDelay for 3 seconds.

```
20240910 215612.746 INFO             PET4  FAIL  Check time delay on non-blocking Component D Finalize/wait, ESMF_CompTunnelUTest.F90, line 1283:  Incorrect time delay
20240910 215612.746 INFO             PET4  delayTime (wait) =    2.9923109999999973
```

https://github.com/esmf-org/esmf/blob/9379ba6a7ea05a870c3c320be556aa76b23df470/src/Superstructure/Component/tests/ESMF_CompTunnelUTest.F90#L1255

https://github.com/esmf-org/esmf/blob/9379ba6a7ea05a870c3c320be556aa76b23df470/src/Superstructure/Component/tests/ESMF_CompTunnelUTest.F90#L1267

https://github.com/esmf-org/esmf/blob/9379ba6a7ea05a870c3c320be556aa76b23df470/src/Superstructure/Component/tests/ESMF_CompTunnelUTest.F90#L1283

_Originally posted by @danrosen25 in https://github.com/esmf-org/esmf/issues/294#issuecomment-2342192503_

danrosen25:
Hi @theurich 
I opened an issue for this. After looking through all the code I couldn't determine the actual point of failure. It could be an issue with the Clang v15 + gcc v14 + OpenMPI v5 call to MPI_Wtime, which could inconsistently return a wall time. This may be the reason precision exists in the first place. wtimedelay shouldn't even return until `wtime@2-wtime@1>=3 seconds` so another call to wtime should guarantee that wtime@3 is greater than wtime@2. It's weird that it doesn't.

I don't have a fix for this but I wanted to note the issue and ask for suggestions.

One more thought is that I don't like the busy wait in `VMK::wtimedelay`. It's burning computational cycles while waiting. The only reason I can think of using this over `std::this_thread::sleep_for(std::chrono::milliseconds(int(delay * 1000)));` is because you don't want to lose anything in the cached memory?

anntsay:
[...]

PR #298, opened 2024-09-17
Fix sprintf issues

billsacks:
Fixes uses of sprintf where the output string is also used in the input, which can result in garbled time strings.

Also changes these sprintf calls to snprintf to avoid possible buffer overflows, and checks these calls to ensure that the output variable is large enough to hold the relevant string, returning an error code if not.

Resolves esmf-org/esmf#284

### DIFF
--- a/src/Infrastructure/Mesh/src/Legacy/ESMCI_MeshMerge.C
+++ b/src/Infrastructure/Mesh/src/Legacy/ESMCI_MeshMerge.C
@@ -1178,12 +1178,12 @@ void concat_meshes(const Mesh & srcmesh, const Mesh & dstmesh, Mesh & mergemesh,
                       for(int npt=0; npt<clip_num_nodes; npt++) std::cout << clip_cd_sph[npt*2] << "  " << clip_cd_sph[npt*2+1] << std::endl; 
                       delete[] clip_cd_sph; 
                     }
-                    char msg[1024];
-                    sprintf(msg," - there was a problem (e.g. repeated points, clockwise poly, etc.) with the triangulation of the element:\n"); 
-                    sprintf(msg,"%selem Id: %d clip_elem Id: %d\n", msg, elem.get_id(), clip_elem.get_id());
+                    std::stringstream msg;
+                    msg << " - there was a problem (e.g. repeated points, clockwise poly, etc.) with the triangulation of the element:\n";
+                    msg << "elem Id: " << elem.get_id() << " clip_elem Id: " << clip_elem.get_id() << "\n";
                     {
-                      cd_sph = new double[num_p*2];   cart2sph(num_p, pts, cd_sph); 
-                      for(int npt=0; npt<num_p*2; npt++) sprintf(msg,"%s %g", msg, cd_sph[npt]); 
+                      cd_sph = new double[num_p*2];   cart2sph(num_p, pts, cd_sph);
+                      for(int npt=0; npt<num_p*2; npt++) msg << " " << cd_sph[npt];
                       delete[] cd_sph; 
                     }
                     if (ESMC_LogDefault.MsgFoundError(ESMC_RC_ARG_INCOMP, msg,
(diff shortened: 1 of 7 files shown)


billsacks:
@oehmke - I'd be happy for a review of this if you'd like to look it over - but I feel good enough about these changes that I don't think a review is necessary if you don't have the time.

src/Infrastructure/Mesh/src/Legacy/ESMCI_MeshMerge.C
```diff
@@ -1178,12 +1178,12 @@ void concat_meshes(const Mesh & srcmesh, const Mesh & dstmesh, Mesh & mergemesh,
...
-                    sprintf(msg," - there was a problem (e.g. repeated points, clockwise poly, etc.) with the triangulation of the element:\n"); 
-                    sprintf(msg,"%selem Id: %d clip_elem Id: %d\n", msg, elem.get_id(), clip_elem.get_id());
+                    std::stringstream msg;
+                    msg << " - there was a problem (e.g. repeated points, clockwise poly, etc.) with the triangulation of the element:\n";
+                    msg << "elem Id: " << elem.get_id() << " clip_elem Id: " << clip_elem.get_id() << "\n";
                     {
-                      cd_sph = new double[num_p*2];   cart2sph(num_p, pts, cd_sph); 
-                      for(int npt=0; npt<num_p*2; npt++) sprintf(msg,"%s %g", msg, cd_sph[npt]); 
+                      cd_sph = new double[num_p*2];   cart2sph(num_p, pts, cd_sph);
+                      for(int npt=0; npt<num_p*2; npt++) msg << " " << cd_sph[npt];
```
billsacks, line 1186: Note that this change is untested because it is inside an `#if 0` block.

I tested it by:

(1) Testing the build when I changed `#if 0` to `#if 1` in the surrounding block

(2) Writing the following test program that checks the msg string before and after this change, ensuring that the resulting string is identical before and after:

```C++
#include <iostream>
#include <sstream>

using namespace std;

int main() {
   int num_p = 4;
   double *cd_sph = new double[num_p * 2];
   for (int npt = 0; npt < num_p * 2; npt++) {
      cd_sph[npt] = 24 + 36.003*npt;
   }

   char msg[1024];
   sprintf(msg, " - there was a problem (e.g. repeated points, clockwise poly, etc.) with the triangulation of the element:\n");
   sprintf(msg, "%selem Id: %d clip_elem Id: %d\n", msg, 1005, 123456);
   for (int npt = 0; npt < num_p * 2; npt++)
      sprintf(msg, "%s %g", msg, cd_sph[npt]);

   std::stringstream msg2;
   msg2 << " - there was a problem (e.g. repeated points, clockwise poly, etc.) with the triangulation of the element:\n";
   msg2 << "elem Id: " << 1005 << " clip_elem Id: " << 123456 << "\n";
   for (int npt = 0; npt < num_p * 2; npt++)
      msg2 << " " << cd_sph[npt];

   cout << msg << "\n";
   cout << msg2.str() << "\n";
}
```

billsacks:
It looks like the test failure here is just in the ESMF_CompTunnelUTest, which I think @danrosen25 has seen failing in these automated runs periodically.

danrosen25:
> It looks like the test failure here is just in the ESMF_CompTunnelUTest, which I think @danrosen25 has seen failing in these automated runs periodically.

I checked and it's the same error about Finalize finishing in less than 3 seconds even though there's a 3 second delay. The issue for it is here: https://github.com/esmf-org/esmf/issues/295

danrosen25:
@billsacks 
I reran the failed mac os + openmpi test after rolling back the openmpi version. This time it passed. It could be a bug in openmpi or it could just be that openmpi v5+ calls to MPI_Wtime execute faster.

oehmke (approved):
Looks great! I appreciate your careful testing, especially of the edge cases. It's nice to get this in. It'll make the library safer. Thanks!

PR #299, opened 2024-09-18
Roll back openmpi version for dev tests

danrosen25:
The `macos@14-clang-gfortran@14-openmpi-netcdf` development test has been failing as noted in 
https://github.com/esmf-org/esmf/issues/295

Attempting to use OpenMPI v4.x.x

Existing cache must be deleted


danrosen25:
@theurich @billsacks 
I've had three tests complete successfully with this OpenMPI version ... so we'll see!

PR #306, opened 2024-10-12
Fix FindPython settings for ESMX

### LINKED ISSUE #305: Python yaml module in ESMX build
There has been an issue in a few places where the ESMX build grabs a different python than one that's already in use and that python doesn't have pyyaml installed so it fails. 

One example of this was in our testing on Orion where our python/yaml based test system was working, but then the ESMX nuopc protos failed because they grabbed a different Python that didn't have yaml. 

Also, one of our collaborators on the SWORD project was having this issue (see email below). (@uturuncoglu  and @anntsay are also interested in this issue from that project.) 

Obviously, one solution is just to install yaml in the python that CMake is grabbing. However, I wonder if it would be possible to have CMake just use the default/current python if one exists, so this doesn't have to be done. 

Thanks!

===== Email =====

Hi Ufuk,

So I installed pyyaml with

pip install pyyaml

and if I try to run the script interactively

python /Users/gtoth/ESMF_nagfor/include/ESMX/Driver/esmx_app_config.py -h

it works fine. If I change the

import yaml

line in  /Users/gtoth/ESMF_nagfor/include/ESMX/Driver/esmx_tools.py to

import yamlxx

then it fails

 File "/Users/gtoth/ESMF_nagfor/include/ESMX/Driver/esmx_tools.py", line 4, in <module>
   import yamlxx
ModuleNotFoundError: No module named 'yamlxx'

so clearly, I have yaml in my Python distribution. Still, if I execute

make test_esmf

it fails with

/Users/gtoth/ESMF_nagfor/bin/bing/Darwin.nag.64.mpich.default/ESMX_Builder --prefix=/Users/gtoth/Tmp_nag/SWMF --verbose
Settings:

 ESMX_BUILD_FILE=esmxBuild.yaml
 ESMF_ESMXDIR=/Users/gtoth/ESMF_nagfor/include/ESMX
 ESMFMKFILE=/Users/gtoth/ESMF_nagfor/esmf.mk
 BUILD_DIR=/Users/gtoth/Tmp_nag/SWMF/ESMF/ESMF_SWMF/src/build
 INSTALL_PREFIX=/Users/gtoth/Tmp_nag/SWMF
 BUILD_TYPE=release
 BUILD_ARGS=
 DISABLE_COMPS=
 BUILD_JOBS=
 MODULEFILE=
 BASHENV=
 TEST=false
 TEST_ARGS=
 VERBOSE=true

-- Detected ESMF Beta snapshot: 0
-- Detected ESMF Beta snapshot: 0
Traceback (most recent call last):
 File "/Users/gtoth/ESMF_nagfor/include/ESMX/Driver/esmx_app_config.py", line 4, in <module>
   from esmx_tools import *
 File "/Users/gtoth/ESMF_nagfor/include/ESMX/Driver/esmx_tools.py", line 4, in <module>
   import yaml
ModuleNotFoundError: No module named 'yaml'
CMake Error at Driver/CMakeLists.txt:12 (message):
 esmx_app_config.py failed processing esmxBuild.yaml
Call Stack (most recent call first):
 Driver/CMakeLists.txt:34 (esmx_check_ret)

so in this case "yaml" is not found. Since I do the make test_esmf as myself, I don't understand how it is succeeding to find yaml in on case and failing in the other. Are you potentially using a fixed python version? For example, if I switch from 3.11 to 3.10 I get

<mesh:~/Tmp_nag/SWMF>python3.10 /Users/gtoth/ESMF_nagfor/include/ESMX/Driver/esmx_app_config.py -h
Traceback (most recent call last):
 File "/Users/gtoth/ESMF_nagfor/include/ESMX/Driver/esmx_app_config.py", line 4, in <module>
   from esmx_tools import *
 File "/Users/gtoth/ESMF_nagfor/include/ESMX/Driver/esmx_tools.py", line 4, in <module>
   import yaml
ModuleNotFoundError: No module named 'yaml'

Clearly, it would be a good idea to use the default python version.

Gabor

danrosen25:
Fixes #305
Find the most generic Python interpreter in standard paths

* skip frameworks
* use unversioned Python interpreter names

### DIFF
--- a/src/addon/ESMX/Driver/CMakeLists.txt
+++ b/src/addon/ESMX/Driver/CMakeLists.txt
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 3.22)
 
 # load packages and modules
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_LIST_DIR}/cmake)
+set(Python_FIND_UNVERSIONED_NAMES "FIRST")
+set(Python_FIND_FRAMEWORK "LAST")
+set(Python_FIND_REGISTRY  "LAST")
 find_package(Python 3.5 REQUIRED COMPONENTS Interpreter)
 find_package(ESMF 8.6.0 MODULE REQUIRED)
 include(FetchContent)
--- a/src/addon/ESMX/README.md
+++ b/src/addon/ESMX/README.md
@@ -41,6 +41,7 @@ options:
   [--disable-comps=DISABLE_COMPS]
   [--build-type=BUILD_TYPE] or [-g]
   [--build-args=BUILD_ARGS]
+  [--cmake-args=CMAKE_ARGS]
   [--build-jobs=JOBS]
   [--load-modulefile=MODULEFILE]
   [--load-bashenv=BASHENV]
@@ -68,7 +69,10 @@ where:
   --build-type=BUILD_TYPE  build type; valid options are 'debug', 'release', 'relWithDebInfo'
   -g                       (default: release) (-g sets BUILD_TYPE to debug)
 
-  --build-args=BUILD_ARGS  global cmake arguments (e.g. -DVAR=val)
+  --build-args=BUILD_ARGS  build arguments passed to every component (e.g. -DVAR=val)
+
+  --cmake-args=CMAKE_ARGS  cmake arguments are used to configure cmake while building ESMX
+                           (e.g. -DVAR=val)
 
   --build-jobs=BUILD_JOBS  number of jobs used for building esmx and components
 
--- a/src/apps/ESMX_Builder/ESMX_Builder.sh
+++ b/src/apps/ESMX_Builder/ESMX_Builder.sh
@@ -29,7 +29,14 @@ usage () {
   printf "  -g\n"
   printf "      set --build-type=debug\n"
   printf "  --build-args=BUILD_ARGS\n"
-  printf "      cmake arguments (e.g. -DARG=VAL)\n"
+  printf "      build args are passed to each component\n"
+  printf "      arguments are not used to configure ESMX (see --cmake-args)\n"
+  printf "      (e.g. --build-args=\"-DFOO=BAR\")\n"
+  printf "      note: build_args in ESMX_BUILD_FILE supersedes --build-args\n"
+  printf "  --cmake-args=CMAKE_ARGS\n"
+  printf "      cmake args are used to configure cmake while building ESMX\n"
+  printf "      arguments are not passed to components (see --build-args)\n"
+  printf "      (e.g. --cmake-args=\"-DFOO=BAR\")\n"
   printf "  --disable-comps=DISABLE_COMPS\n"
   printf "      disable components\n"
   printf "  --build-jobs=BUILD_JOBS\n"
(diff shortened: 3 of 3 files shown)


danrosen25:
@uturuncoglu @oehmke 
I think these changes will lead the ESMX build infrastructure to prioritize the version of python interpreter typically called by users. The default for FindPython.cmake seems to prioritize a framework.

oehmke (approved):
I haven't done a lot of CMake, so it's hard for me to tell exactly what this will do. However, if you think that this'll be more likely to grab the user's version/the default version, then it sounds good to me.   Once we get it merged we can have Gabor try the next snapshot.  Thanks for doing this so quickly!

uturuncoglu:
@danrosen25 since this is draft. I did not check carefully. I wonder if we could those changes optional. Is it possible to introduce a an option when we are calling ESMX build. So, user might have flexibility to chose the desired way. We could also make the existing default doe it would be backward compatible.

danrosen25:
> @danrosen25 since this is draft. I did not check carefully. I wonder if we could those changes optional. Is it possible to introduce a an option when we are calling ESMX build. So, user might have flexibility to chose the desired way. We could also make the existing default doe it would be backward compatible.

@uturuncoglu - Not organically. The build needs to find Python (with pyyaml) before processing the build options. There's a way to use CMake variables to control ESMX build options. I added an option for `--cmake-args="-DFOO=BAR"` to the ESMX_Builder script. This allows users to pass something like `--cmake-args="-DPython_EXECUTABLE=/opt/homebrew/bin/python3.11"`. I left the default to find the un-versioned/non-framework Python version. Let me know if you like this better?

oehmke:
So basically by default it will try to find the generic standard python version that the user has, but if they need to then they can force it to use a specifc python version? If so, I like that.

uturuncoglu:
@danrosen25 Yes. That is great. Thanks. BTW, user will have flexibility.

uturuncoglu:
@oehmke I wonder if we need to test this version of ESMF in Gabor case once it is in develop to see how it goes.

uturuncoglu (approved):

oehmke:
Yep, that makes sense to me. Once it’s in, I can put a snapshot on, and we can give it to him to try.

uturuncoglu:
@oehmke I think he also success since he sand a mail indicates that after installing pyyaml into python 3.9 (which is _not_ the default python version on his machine), he could able to compile the code.

danrosen25:
@theurich 
Is this wording better?
```
  --build-args=BUILD_ARGS
      build args are passed to each component
      arguments are not used to configure ESMX (see --cmake-args)
      (e.g. --build-args="-DFOO=BAR")
      note: build_args in ESMX_BUILD_FILE supersedes --build-args
  --cmake-args=CMAKE_ARGS
      cmake args are used to configure cmake while building ESMX
      arguments are not passed to components (see --build-args)
      (e.g. --cmake-args="-DFOO=BAR")
```

theurich:
> @theurich Is this wording better?
> 
> ```
>   --build-args=BUILD_ARGS
>       build args are passed to each component
>       arguments are not used to configure ESMX (see --cmake-args)
>       (e.g. --build-args="-DFOO=BAR")
>       note: build_args in ESMX_BUILD_FILE supersedes --build-args
>   --cmake-args=CMAKE_ARGS
>       cmake args are used to configure cmake while building ESMX
>       arguments are not passed to components (see --build-args)
>       (e.g. --cmake-args="-DFOO=BAR")
> ```

Yes, that looks good to me. 
Question: is there also `cmake_args` available as an option under the `application` section of the ESMX_BUILD_FILE?

danrosen25:
@theurich 
There isn't an equivalent in esmxBuild.yaml. The scope for options in esmxBuild.yaml is more limited than ESMX_Builder. We can do something similar but not exactly the same as the scope will be different.

theurich:
I am find with how it is now. Was just curious.

Issue #293, opened 2024-09-09
Floating point lists for NUOPC_CompAttributeGet


danrosen25:
### Discussed in https://github.com/orgs/esmf-org/discussions/291

Using ESMF_Config with a list of items isn't working unless items do not have whitespace and are parsed in user code. Reproducer attached

**What's broken**
```
<COMP>_attributes::
      restart_fh = 0.05,0.25,0.5,3,6
 ::
```
Lists aren't working. This is seen as a single string.
```
<COMP>_attributes::
      restart_fh = 0.05, 0.25, 0.5, 3, 6
 ::
```
Spaces are breaking ESMF_Info. So NUOPC attributes doesn't store `restart_fh`at all. This is also true for `restart_fh = "0.05, 0.25, 0.5, 3, 6"`.
[reproducer_info.tar.gz](https://github.com/user-attachments/files/16934744/reproducer_info.tar.gz)

anntsay:
punt this to 8.9.. as we may consider a different implementation.  for 8.8 will add documentation change.

PR #316, opened 2024-10-31
Fix issue #293 NUOPC_CompAttributeIngest failure

danrosen25:
Append a token lists into single spaced string


theurich (approved):
This looks good to me. The way I interpret the new code, the use ends up with a single string value for both cases:
```
<COMP>_attributes::
      restart_fh = 0.05,0.25,0.5,3,6
 ::
```
 and
```
<COMP>_attributes::
      restart_fh = 0.05, 0.25, 0.5, 3, 6
 ::
```
I think that is consistent behavior for how the non-HConfig Attribute handling is working string based.

danrosen25:
@theurich It does but note that
```
<COMP>_attributes::
      restart_fh = 0.05, 0.25, 0.5, 3, 6
 ::
```
and
```
<COMP>_attributes::
      restart_fh = 0.05,     0.25,          0.5, 3,  6
 ::
```
both return `0.05, 0.25, 0.5, 3, 6`

So although it works with whitespace now
```
<COMP>_attributes::
      my_line = I  am  a  sentence  with  double  spacing
 ::
```
returns `I am a sentence with double spacing`
and that's not what a user would expect

I had to make a change because the NUOPC Comp UTest was failing.

danrosen25:
@tclune @NickSzapiro-NOAA
Hi,
I'm looking for user feedback for this change to NUOPC. It's become standard practice to read ESMF Config tables into a FreeFormat then ingest the FreeFormat into component attributes. We discovered a bug in which ESMF Config table rows that include spaces were being completely ignored `restart_fh = 0.05, 0.25, 0.5, 3, 6` during this process. This happens because reading config into a FreeFormat treats spaces as column separators, which is how we support tables. When ingesting a FreeFormat into an attribute NUOPC does not know how to handle multiple attribute values for a single key.

This fix will concatenate all of the items with one space between each item, which is then added to the component's attributes. It's not without issues because it will trim all excess white space. See code for specific technical details.

I did play around with ingesting multiple values as a valueList but there was little benefit.

tclune:
@danrosen25 I doubt this will affect GEOS, but if you want us to test a branch with GEOS, we'll be happy to oblige.

Wish I could say we are entirely HConfig at this point, but production systems will be mostly Config for at least another year, not to mention legacy variants.

tclune:
Sorry - was assuming this was a change in the underlying ESMF_Config.   GEOS itself does not use any NUOPC code, so this should only affect UFS.

anntsay:
this is a new feature.. current status quo is that you get nothing returned.. this new implementation, it will at least return it as a single space list.  and the user can do further processing themselves.

However, we also should let the user determine how they want to use this.. therefore, team decide to provide better documentation on how config handle white spaces 

Action:
1. beef up documentation and caution user on the behavior of list and white space.
2. punt this fix for now in 8.8.0.. maybe reconsider other implemenation

anntsay:
@danrosen25  will close this PR and consider different implementation for #293

PR #324, opened 2024-11-22
Make `status` optional in ESMF_FieldEmptyReset()

theurich:
@oehmke the new ESMF_FieldEmptyReset() looks good to me. Only thing I was thinking would be nice is to make the `status` argument optional. This is what this PR does.


oehmke (approved):
Thanks for reviewing! The changes that you suggest look good to me. I like the idea of making status optional.

Issue #69, opened 2022-09-08
Complete time manager & alarm fixes


rsdunlapiv:

theurich:
@rsdunlapiv do you remember what branch is associated with this issue? Is it https://github.com/esmf-org/esmf/tree/newalarm? Thanks.

billsacks:
See also https://docs.google.com/document/d/1ZlI12Q8XPwQgPCuu1feyrUorhfuxkxhQMG50Iv97jU8/edit - sent by Arun Sept 14, 2023 with the note:

> One key requirement as we look forward to Data Assimilation is being able to move the model forward and backward which will have some work to be done on the ESMF clock side of things. I am forwarding you a requirements document with the kinds of tests we would like to see the model pass.

anntsay:
Other related  alarm tickets:  https://github.com/esmf-org/esmf-support/issues/11, https://github.com/esmf-org/esmf-support/issues/15

anntsay:
recap meeting.. may need to update the branch to 8.7.

anntsay:
Dan said that Fei had tried it but we need to test out other places/systems.  Action - give UFS team the branch to try out time management. (Ann to follow up in the next UFS meeting).

anntsay:
That this is NOT backward compatible and therefore need to move the branch to 9.0

billsacks:
> That this is NOT backward compatible and therefore need to move the branch to 9.0

After a bit more discussion, we're not sure about this point. The key question is whether alarms are still sticky by default. Our initial thinking was that alarm stickiness is no longer supported, which would break backwards compatibility, but from further discussion, it seems like we're not sure about that point and it needs some further investigation.

anntsay:
Action: 1. look at the original issue.
            2. design based

Consideration: has a new alarm class to try it out.  but may also need to consider clock.

Impact:  NASA is still asking about this.

anntsay:
NASA has a workaround for their alarm class. possibly to leverage that and start building new alarm class base.

anntsay:
Bob currently cross-referencing with Tom's manifesto

PR #340, opened 2024-12-26
Newalarm

billsacks:
Replaces #121 .

Relevant issues:
- https://github.com/esmf-org/esmf/issues/69
- https://github.com/esmf-org/esmf-support/issues/11
- https://github.com/esmf-org/esmf-support/issues/15
- https://github.com/esmf-org/esmf-support/issues/89

See also:
- https://github.com/esmf-org/esmf/pull/29


billsacks:
I found what I think is an instance where the cleanup on this branch has broken some behavior, and this makes me nervous about what else might have gotten broken by the cleanup here.

The case that I think is now broken involves:
- Setting a one-shot (i.e., ringTime but no ringInterval) sticky alarm.
- After running the clock forward for some time, setting its direction to reverse
- Turning off the alarm, possibly multiple times

What I noticed was that, each time the alarm is turned off, its ringTime is adjusted by one time step. It looks like this is coming from the `ringerOff` function, which has this code:

```C++
    if (clock->direction == ESMF_DIRECTION_FORWARD) {
      // remember this time, so if we go in reverse, we will know when
      //   to turn the alarm back on.
      //   TODO:  Assumes constant ringInterval between successive ringEnds;
      //          saves only last ringEnd.  Make ringEnd an array to save all
      //          end times, which may vary (e.g. due to variable timeSteps).
      ringEnd = clock->currTime;

    } else {      // ESMF_DIRECTION_REVERSE
      // for sticky alarms, step back ring times
      if (sticky && ringTime != firstRingTime) {
        ringTime     -= ringInterval;
        prevRingTime -= ringInterval; 

        // get clock's timestep direction: positive or negative
        bool positive =
               (clock->currAdvanceTimeStep.absValue() ==
                clock->currAdvanceTimeStep) ? true : false;

        // step back ringEnd only if it was advanced past ringTime
        //  before the clock loop ended (covers case where last ringTime
        //  equals the clock->stopTime and alarm is processed before
        //  the clockAdvance()).
        if ((positive  && ringEnd > ringTime) ||
            (!positive && ringEnd < ringTime)) {
          ringEnd -= ringInterval;
        }
      }
    }
```

This code – or something like it – is also on develop, but there are a few differences between this branch and develop that lead this to be a problem on this branch whereas it doesn't appear to be a problem on develop:

First, on this branch, there is code that sets ringInterval to the negative time step if it isn't set initially. (On develop, it appears that ringInterval remains 0 if it isn't set.):

```C++
      // A negative ringInterval indicates a one shot alarm at ringTime only
      TimeInterval zeroTimeInterval(0,0,1,0,0,0);
      if(clock->timeStep < zeroTimeInterval)
        alarm->ringInterval = clock->timeStep;
      alarm->ringInterval = -clock->timeStep;
```

Second, on this branch, it appears that `firstRingTime` is never set. (On develop, it looks like we never get into this conditional that adjusts ringTime for a one-shot sticky alarm, presumably because `ringTime == firstRingTime` (though I haven't confirmed that)... though even if we did, the adjustment would be by 0 because ringInterval is 0.)

So here's a task for this PR:
- [ ] Fix this issue with ringTime being adjusted by a time step whenever setting an alarm to off.

src/Infrastructure/TimeMgr/src/ESMCI_Alarm.C
```diff
@@ -129,59 +128,61 @@ int Alarm::count=0;
...
+      // LCM albeit it may be very large.
+      alarm->ringInterval = *ringInterval;
+
+    }else{
+      // A negative ringInterval indicates a one shot alarm at ringTime only
+      TimeInterval zeroTimeInterval(0,0,1,0,0,0);
+      if(clock->timeStep < zeroTimeInterval)
+        alarm->ringInterval = clock->timeStep;
+      alarm->ringInterval = -clock->timeStep;
     }
```
billsacks, line 173: - [ ] This seems problematic, particularly given the code in `ringerOff` that decrements `ringTime` by `ringInterval`. (See also https://github.com/esmf-org/esmf/pull/340#issuecomment-2566736806.) I think it would be better and safer to have a boolean flag that says whether this alarm rings on an interval or is a one-shot alarm.

src/Infrastructure/TimeMgr/src/ESMCI_Alarm.C
```diff
@@ -129,59 +128,61 @@ int Alarm::count=0;
       sprintf(alarm->name, "Alarm%3.3d", alarm->id);
     }
 
-    if (ringTime != ESMC_NULL_POINTER) {
-      alarm->ringTime = alarm->prevRingTime = alarm->firstRingTime = *ringTime;
```
billsacks, line 133: - [ ] I think it's a problem that `firstRingTime` is no longer set: it's still referenced in `ringerOff` (see https://github.com/esmf-org/esmf/pull/340#issuecomment-2566736806)

src/Infrastructure/TimeMgr/src/ESMCI_Alarm.C
```diff
@@ -438,6 +436,12 @@ int Alarm::count=0;
 
  } // end Alarm::set
 
+void Alarm::enableSticky(void){
+  // 1/100th of a second granuality to ensure integer return when being divided by
+  //this->ringInterval=TimeInterval(0, 1, 100, 0, 0, 0); 
+  this->ringInterval=TimeInterval(1, 0, 1, 0, 0, 0); 
+}
```
billsacks, line 443: - [ ] This warrants investigation: First, note that this is now using a 1 second granularity rather than the 1/100 second mentioned in the comment. But more importantly, it looks like this setting of `ringInterval` will differ depending on whether an alarm is initially created as sticky (in which case I don't think `enableSticky` is executed, and so ringInterval is either kept at what is set or set to a negative value for a one-shot), or if it's later set as sticky (in which case this `enableSticky` is called).

src/Infrastructure/TimeMgr/src/ESMCI_Alarm.C
```diff
@@ -1117,320 +1128,85 @@ int Alarm::count=0;
...
-    return(ringing && enabled);
+  // If the direction didn't change, do nothing
+  if(old_direction == new_direction ) return;
 
- } // end Alarm::checkRingTime
+  if(old_direction == ESMF_DIRECTION_FORWARD) 
+    ringTime = ringTime - ringInterval; // ringTime goes backward by 1 interval
+  else
+    ringTime = ringTime + ringInterval; // ringTime goes forward by 1 interval
+}
```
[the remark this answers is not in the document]
billsacks: Note, though, that this is at least somewhat similar to what's done on develop if `userChangedRingTime` is true.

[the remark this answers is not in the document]
billsacks: Actually, I think my initial understanding of why this is done was wrong:

I think I was thinking about this in terms of a one-shot alarm, and I actually think **we probably don't want to apply this in the case of a one-shot alarm** (this is currently problematic because of the way `ringInterval` is set for a one-shot alarm; it should probably either be 0 or we should have an explicit flag saying if this is a one-shot alarm, and if so, should avoid doing this adjustment).

But I actually see the rationale for why this is done for a repeating alarm: If `ringTime` gives the next time the alarm will ring, then we reverse directions in between two rings, then we need to adjust `ringTime` by one `ringInterval` so that it gives the "next time" in this new direction.

So maybe we want this in place, but we want to avoid triggering it in some cases: for one-shot alarms and *maybe* some other cases (e.g., what if `ringTime` is already exactly at the current time?).

src/Infrastructure/TimeMgr/tests/ESMF_AlarmUTest.F90
```diff
@@ -3598,9 +1606,1370 @@ subroutine Test_AlarmHang(testPass, rc)
...
+      ! in this case, this test returns SUCCESS
+      rc = ESMF_SUCCESS
+    else
+      ! On the other hand, if ESMF_AlarmCreate succeeded, then this test has failed
+      rc = ESMF_FAILURE
+    end if
+    !call ESMF_AlarmDebug(alarm,'test3',rc=status)
+  end subroutine
+!------------------------------------------------------------------------
+  subroutine ForwardAlarm_Test4(rc)
```
billsacks, line 2128: - [ ] The current behavior as demonstrated in this test seems problematic and unintuitive, especially for sticky alarms: As demonstrated here, when the alarm dt doesn't align with the clock dt, the alarm will only ring when the two come into alignment, even for a sticky alarm. This is (I think) in contrast to the previous behavior, when the alarm would ring when we first pass the ring time.

anntsay:
Need to review this PR against the new proposal from the NASA alarm branch.

PR #342, opened 2025-01-08
Fix clock drift macOS error in ESMF_CompTunnelUTest

danrosen25:


src/Superstructure/Component/tests/ESMF_CompTunnelUTest.F90
```diff
@@ -18,6 +18,14 @@
 ! The SLEEPTIME macro defines the time in seconds that Finalize will delay
 #define SLEEPTIME 3.0_ESMF_KIND_R8
 
+! The DRIFTTIME macro defines a clock drift tolerance in seconds
+! DRIFTTIME is used to capture system clock errors on macOS runners
```
theurich, line 22:
```suggestion
! DRIFTTIME is used to work around system clock errors on macOS runners
```

theurich (approved):
Approve, with tiny suggestion request in wording of comment.

Issue #346, opened 2025-01-28
Leap seconds


oehmke:
The idea of handling leap seconds came up on the NASA call. They don't need them now, and they don't foresee needing them any time soon. However, they were just wondering if we had thought about them.  It would be good to at least document the fact that we don't consider them in the TimeManger.  Also, while we are looking at changing some functionality in the TimeManager, it might make sense to think about the issue a bit.

billsacks:
I just did a moment's research on this and it frankly sounds like a nightmare. From the [Wikipedia page on Leap seconds](https://en.wikipedia.org/wiki/Leap_second) (see particularly the bolded bit at the end – emphasis mine):

> The leap second was introduced in 1972. Since then, 27 leap seconds have been added to UTC, with the most recent occurring on December 31, 2016.[[1]](https://en.wikipedia.org/wiki/Leap_second#cite_note-Martin-2024-1) All have so far been positive leap seconds, adding a second to a UTC day; while it is possible for a negative leap second to be needed, this has not happened yet.
> 
> Because the Earth's rotational speed varies in response to climatic and geological events,[[2]](https://en.wikipedia.org/wiki/Leap_second#cite_note-2) UTC leap seconds are irregularly spaced and unpredictable. Insertion of each UTC leap second is usually decided about six months in advance by the [International Earth Rotation and Reference Systems Service](https://en.wikipedia.org/wiki/International_Earth_Rotation_and_Reference_Systems_Service) (IERS), to ensure that the difference between the UTC and UT1 readings will never exceed 0.9 seconds.[[3]](https://en.wikipedia.org/wiki/Leap_second#cite_note-Bulletin_C_49-3)[[4]](https://en.wikipedia.org/wiki/Leap_second#cite_note-4)
> 
[...]

PR #347, opened 2025-01-29
Document that ESMF does *not* handle leap seconds

billsacks:
Resolves esmf-org/esmf#346


billsacks:
@oehmke - from a quick look at the documentation, this felt like the most logical place to include this documentation, since I'm thinking that someone interested in leap seconds would be looking at details of how ESMF handles the Gregorian calendar implementation. But I'm open to feedback on this location.

PR #356, opened 2025-02-19
Fix protex to enable it to work with Perl 5.30.

oehmke:
Remove some outdated code that is no longer supported in recent versions of Perl.

### DIFF
--- a/scripts/doc_templates/templates/protex
+++ b/scripts/doc_templates/templates/protex
@@ -92,7 +92,7 @@
 
 # Keep this if you don't know what it does...
 # -------------------------------------------
-  $[ = 1;                 # set array base to 1
+#  $[ = 1;              # set array base to 1 (Removed to maintain compatability with Perl 5.30)
   $, = ' ';               # set output field separator
   $\ = "\n";              # set output record separator
 
@@ -296,8 +296,8 @@ LINE:
 
 #       Straight quote
 #       --------------
-        if ($Fld[1] eq '!QUOTE:') {
-           for ($i = 2; $i <= $#Fld; $i++) {
+        if ($Fld[0] eq '!QUOTE:') {
+           for ($i = 1; $i <= $#Fld - 1; $i++) {
                printf '%s ', $Fld[$i];
 }#         end for
            print " ";
@@ -306,13 +306,13 @@ LINE:
 
 #       Handle optional Title Page and Introduction
 #       -------------------------------------------
-        if ($Fld[1] eq $boi_string) {
+        if ($Fld[0] eq $boi_string) {
            print ' ';
            $intro = 1;
            next LINE;
 }#      end if
 
-        if ($Fld[2] eq '!TITLE:') {
+        if ($Fld[1] eq '!TITLE:') {
            if ( $intro ) {
               shift @Fld;
               shift @Fld;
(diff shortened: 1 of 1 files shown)


scripts/doc_templates/templates/protex
```diff
@@ -92,7 +92,7 @@
 
 # Keep this if you don't know what it does...
 # -------------------------------------------
-  $[ = 1;                 # set array base to 1
+#  $[ = 1;              # set array base to 1 (Removed to maintain compatability with Perl 5.30)
```
billsacks: I feel like it would be more clear just to remove this line, so I have done that.

billsacks:
I'm looking at this....

oehmke:
Thanks!

scripts/doc_templates/templates/protex
```diff
@@ -296,8 +296,8 @@ LINE:
 
 #       Straight quote
 #       --------------
-        if ($Fld[1] eq '!QUOTE:') {
-           for ($i = 2; $i <= $#Fld; $i++) {
+        if ($Fld[0] eq '!QUOTE:') {
+           for ($i = 1; $i <= $#Fld - 1; $i++) {
```
billsacks: I am removing the change you had here (subtracting 1 from `$#Fld`): `$#Fld` gives the index of the last element, and I think the correct thing to do is still to loop until `i` is equal to the index of the last element. (I did a little test and `$#Fld` changes with `$[`, which I think makes it correct to *not* subtract 1 here.)

billsacks (approved):
Thanks for doing this, @oehmke ! I have forgotten most of the Perl I once knew, but I think I've gotten my head back into it enough to give this a decent (albeit not rock-solid, particularly given the obscurity of the `$[` feature) review. I pushed a commit with some additional changes. It would be great if you gave this commit a quick look to make sure I didn't change anything that you deliberately were keeping as is or think shouldn't be changed:

https://github.com/esmf-org/esmf/pull/356/commits/97b1aef91fa2d5be3ddbff12d2373f5a6fe9c47d

See my inline comments below on two of these changes. Some of the changes are in comments, but I figured they should be changed, too, in case someone uncomments them. I also replaced uses of `$[` with `0`, since now `$[` will be 0 and that seems more clear and standard Perl. In addition, I found a few places that seemed like they should have been changed similarly to your other changes but weren't; I'd especially welcome your look at these to make sure they weren't things that you intentionally left as is.

I was least sure about the changes to the string-related things, but from a little test program, the changes to `substr` seem correct; and it looks like the uses of `index` here are okay; and based on a search for `$[` in https://perldoc.perl.org/5.12.4/perlfunc, I don't see mention of this impacting any other functions.

scripts/doc_templates/templates/protex
```diff
@@ -297,7 +296,7 @@ LINE:
 #       Straight quote
 #       --------------
         if ($Fld[0] eq '!QUOTE:') {
-           for ($i = 1; $i <= $#Fld - 1; $i++) {
+           for ($i = 1; $i <= $#Fld; $i++) {
```
danrosen25, line 299: @billsacks 
Did you mean to change this from `<=` to `<` when you dropped the `- 1`

danrosen25, line 299: I read the Perl documentation. `$#array` is the last index of an array. Good catch.

danrosen25:
@billsacks 
Thanks for cleaning this up ⭐

oehmke:
@billsacks thanks for fixing those remaining issues! I took a look and I agree with your changes. I had a question about the $#Fld change, but that was before I saw your explanation below. I’ll merge this once I get clearance from the gold standard. :-)

oehmke:
@danrosen25 are you ok with me merging this? It sounds like you are, but it doesn't look like you officially approved them, so I wanted to double check.

Issue #353, opened 2025-02-14
ESMF doesn't build with netcdf-c 4.9.3


thebaptiste:
Trying to build ESMF 8.8.0 against netcdf-c 4.9.3, I have this error : 
```
/src/layers/layer1_scientific/0050_esmf/build/esmf-8.8.0/src/Infrastructure/IO/PIO/ParallelIO/src/clib/pio_nc.c: In function 'PIOc_def_var_fill':
/src/layers/layer1_scientific/0050_esmf/build/esmf-8.8.0/src/Infrastructure/IO/PIO/ParallelIO/src/clib/pio_nc.c:2433:56: error: '_FillValue' undeclared (first use in this function); did you mean 'NC_FillValue'?
 2433 |                     ierr = nc_put_att(file->fh, varid, _FillValue, xtype, 1, fill_valuep);
      |                                                        ^~~~~~~~~~
      |                                                        NC_FillValue
```
Looking at [netcdf-c release notes](https://github.com/Unidata/netcdf-c/blob/main/RELEASE_NOTES.md), I found this for 4.9.3 rc 2 : 
```
Release Candidate 2 - December 6, 2024

    Note: To avoid a conflict between _FillValue and libc++18, we have introduced a new option, --enable-legacy-macros for autotools and NETCDF_ENABLE_LEGACY_MACROS for cmake. These are turned on by default currently but will be turned off eventually. Developers are encouraged to move away from the FillValue macro and replace it with the new NC_FillValue macro. See https://github.com/Unidata/netcdf-c/issues/2858 for more information.
```
So I think you should replace `_FillValue` by `NC_FillValue` in ESMF code (as suggested in the error message).
May be adding `#define NETCDF_ENABLE_LEGACY_MACROS` before including `netcdf.h` would do the job too.

Regards

billsacks:
Thank you for making us aware of this issue. Based on the error you're getting and the discussion in https://github.com/Unidata/netcdf-c/issues/3017, it looks like this is a broader issue with compatibility of PIO with the latest NetCDF version; as such, I think this will need to be addressed in either NetCDF or PIO. I'm going to close this here for now, but we can revisit it if it seems like something needs to be done in ESMF beyond the PIO changes.

thebaptiste:
[...]

PR #359, opened 2025-02-24
Update PIO from pio2_6_2 to pio2_6_5

billsacks:
I did this update by simply copying the ParallelIO repository (checked
out at pio2_6_5) to here, and then deleting the .git directory. I
confirmed that, prior to this update, there were no diffs from the
ParallelIO repository (at pio2_6_2), so it is safe to do the update in
this simple manner.

Changes here:
- pio2_6_3: Merge pull request #1982 from alperaltuntas/bugfix_debug: fix minor bugs in pio_nc4.c
- pio2_6_4: Bug fix release. Fixes issues with large datasets and with autoconf/cmake configure option.
- pio2_6_5: Updated for compatibility with netcdf-c 4.9.3 and cygwin.

Resolves esmf-org/esmf#353


billsacks:
@jedwards4b @theurich @oehmke - FYI

I have run testing on my Mac and all tests pass. It looks like, at least for some PIO updates in the past, we have run full nightly testing on a branch before merging. The changes here look fairly minimal, though, so my feeling is that that seems unnecessary here. @theurich or @oehmke - any feelings on whether to do that or just merge and check the regular nightly testing the next day?

oehmke:
If the changes are pretty minimal, I’m ok with just merging them and checking them the next day. (Just keep an eye on the 4 platforms worth of tests that automatically run on develop now and make sure they look ok.)

PR #361, opened 2025-03-05
Add Runtime Options for logMsgAbort and abort

### LINKED ISSUE #296: Option to SIGQUIT or throw error during ESMF_Abort
The current method to debug ESMF Errors is to build a back trace using `ESMF_LogSetError` and `rc`. This gives you a limited amount of information about the state at the time of the error. I started investigating throwing a `SIGQUIT` error, which can print a backtrace and dump a core. The core dump can be analyzed to see the state causing the error.

```
diff --git a/src/Infrastructure/VM/src/ESMCI_VMKernel.C b/src/Infrastructure/VM/src/ESMCI_VMKernel.C
index 63b85ad0c3..43c85c5c5c 100644
--- a/src/Infrastructure/VM/src/ESMCI_VMKernel.C
+++ b/src/Infrastructure/VM/src/ESMCI_VMKernel.C
@@ -899,6 +899,7 @@ struct SpawnArg{
 void VMK::abort(){
   // abort default (all MPI) virtual machine
   int finalized;
+  raise (SIGQUIT);
   MPI_Finalized(&finalized);
   if (!finalized)
     MPI_Abort(default_mpi_c, EXIT_FAILURE);
```

danrosen25:
* use ESMF_RUNTIME_ABORT_LOGMSG_TYPES to set logMsgAbort for ESMF logs
* use ESMF_RUNTIME_ABORT_ACTION to call sigquit for ESMF_Abort

Closes #296 

### Testing Option 1 (uses an ESMF_Initialize configuration file)
Initialize ESMF
```fortran
call ESMF_Initialize(configFilenameFromArgNum=1, configFileName="esmf.cfg", &
  defaultDefaultCalKind=ESMF_CALKIND_GREGORIAN, rc=rc)
```

esmf.cfg
```yaml
ESMF_RUNTIME_ABORT_LOGMSG_TYPES: "ESMF_LOGMSG_ERROR" # calls ESMF_Abort if ERROR is reported to ESMF logs
ESMF_RUNTIME_ABORT_ACTION: "SIGABRT" # signals abort if ESMF_Abort is called
```

### Testing Option 2 (uses environment variables)
set environment variables
```bash
export ESMF_RUNTIME_ABORT_LOGMSG_TYPES="ESMF_LOGMSG_ERROR"
export ESMF_RUNTIME_ABORT_ACTION="SIGABRT" # or "SIGQUIT"
```

### Test Application
[esmf_abort.tar.gz](https://github.com/user-attachments/files/19097557/esmf_abort.tar.gz)

### DIFF
--- a/src/Infrastructure/Util/src/ESMF_LogErr.F90
+++ b/src/Infrastructure/Util/src/ESMF_LogErr.F90
@@ -1730,11 +1730,68 @@ end subroutine ESMF_LogGet
 !      \end{description}
 !
 !EOPI
+
+    integer                        :: localrc
+    integer                        :: msgAbortCnt
+    type(ESMF_LogMsg_Flag)         :: msgAbortLst(6)
+    character(ESMF_MAXSTR)         :: envRtAbort
+    integer                        :: idxRtAbort
+
     ! Initialize return code; assume routine not implemented
     if (present(rc)) rc = ESMF_RC_NOT_IMPL
 
     call ESMF_LogOpen(ESMF_LogDefault, filename,  &
-        appendflag=logappendflag, logkindflag=logkindflag, rc=rc)
+      appendflag=logappendflag, logkindflag=logkindflag, rc=localrc)
+    if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
+      ESMF_CONTEXT, rcToReturn=rc)) return
+
+    call c_ESMC_VMGetEnv("ESMF_RUNTIME_ABORT_LOGMSG_TYPES", envRtAbort, localrc)
+    if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
+      ESMF_CONTEXT, rcToReturn=rc)) return
+    msgAbortCnt = 0
+    idxRtAbort = index (envRtAbort, "ESMF_LOGMSG_DEBUG")
+    if (idxRtAbort > 0) then
+      msgAbortCnt = msgAbortCnt + 1
+      msgAbortLst(msgAbortCnt) = ESMF_LOGMSG_DEBUG
+    endif
+    idxRtAbort = index (envRtAbort, "ESMF_LOGMSG_ERROR")
+    if (idxRtAbort > 0) then
+      msgAbortCnt = msgAbortCnt + 1
+      msgAbortLst(msgAbortCnt) = ESMF_LOGMSG_ERROR
+    endif
+    idxRtAbort = index (envRtAbort, "ESMF_LOGMSG_INFO")
+    if (idxRtAbort > 0) then
+      msgAbortCnt = msgAbortCnt + 1
+      msgAbortLst(msgAbortCnt) = ESMF_LOGMSG_INFO
+    endif
+    idxRtAbort = index (envRtAbort, "ESMF_LOGMSG_JSON")
+    if (idxRtAbort > 0) then
+      msgAbortCnt = msgAbortCnt + 1
+      msgAbortLst(msgAbortCnt) = ESMF_LOGMSG_JSON
+    endif
+    idxRtAbort = index (envRtAbort, "ESMF_LOGMSG_TRACE")
+    if (idxRtAbort > 0) then
+      msgAbortCnt = msgAbortCnt + 1
+      msgAbortLst(msgAbortCnt) = ESMF_LOGMSG_TRACE
+    endif
+    idxRtAbort = index (envRtAbort, "ESMF_LOGMSG_WARNING")
+    if (idxRtAbort > 0) then
+      msgAbortCnt = msgAbortCnt + 1
+      msgAbortLst(msgAbortCnt) = ESMF_LOGMSG_WARNING
+    endif
+    if (msgAbortCnt > 6) then
+      call ESMF_LogSetError(ESMF_RC_BUFFER_SHORT, &
+        msg="msgAbortLst overflow - ESMF_RUNTIME_ABORT_LOGMSG_TYPES", &
+        ESMF_CONTEXT, rcToReturn=rc)
+      return
+    elseif (msgAbortCnt > 0) then
+      call ESMF_LogSet(ESMF_LogDefault, &
+        logmsgAbort=msgAbortLst(1:msgAbortCnt), rc=localrc)
+      if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
+        ESMF_CONTEXT, rcToReturn=rc)) return
+    endif
+
+    if (present(rc)) rc=ESMF_SUCCESS
 
 end subroutine ESMF_LogInitialize
 
--- a/src/addon/ESMX/README.md
+++ b/src/addon/ESMX/README.md
@@ -344,6 +344,8 @@ This section affects the application level.
 | `defaultCalKind`          | ESMF calendar kind used by default, see ESMF RefDoc for options | `ESMF_CALKIND_GREGORIAN`|
 | `logFlush`                | enable/disable log flush for each write: `true` or `false`| `false`         |
 | `fieldDictionary`         | name of the NUOPC field dictionary file to be loaded      | *None*          |
+| `ESMF_RUNTIME_ABORT_ACTION`| set ESMF abort action: `MPI_ABORT` `SIGABRT` or `SIGQUIT`| `MPI_ABORT`     |
+| `ESMF_RUNTIME_ABORT_LOGMSG_TYPES`| type(s) of log messages causing abort (e.g. `ESMF_LOGMSG_ERROR,ESMF_LOGMSG_WARNING`) | *None* |
 | `ESMF_RUNTIME_COMPLIANCECHECK`| enable/disable NUOPC compliance checking: `ON` or `OFF` with `DEPTH` | `OFF`|
 | `ESMF_RUNTIME_GARBAGE`    | ESMF garbage collection setting: `NONE`, `FULL`,`SAFE`     | `SAFE`          |
 | `ESMF_RUNTIME_GARBAGE_LOG`| enable/disable logging of ESMF garbage collection: `ON` or `OFF`         | `OFF`|
--- a/src/doc/ESMF_debug.tex
+++ b/src/doc/ESMF_debug.tex
@@ -155,3 +155,27 @@ is a problem in the library code. In this example user code is visible at the
 very top of the stack {\tt esmApp.F90}, and at the very bottom {\tt ocn.F90}.
 In fact the location of the division by zero is correctly identified by the
 runtime library at line 149 in file {\tt ocn.F90}.
+
+It is possible to force a hard crash within the ESMF library while logging
+ERRORS or WARNINGS to the {\tt PET<nnn>.ESMF\_LogFile}. Doing so can be
+advantageous because it may produce a code dump and/or backtrace at the
+initial point of error without ESMF return code handling. There are two
+settings controlling log message error handling. The first setting,
+{\tt ESMF\_RUNTIME\_ABORT\_LOGMSG\_TYPES}, configures the ESMF library to abort
+during specified log message types, such as {\tt ESMF\_LOGMSG\_ERROR}. When
+setting {\tt ESMF\_RUNTIME\_ABORT\_LOGMSG\_TYPES}, multiple log message types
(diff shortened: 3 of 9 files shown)


danrosen25:
I spoke with the team and they like this feature as an alternative method for debugging. It extends the capability of setting logMsgAbort for ESMF logs to a runtime configurable option. It also extents the capability of ESMF_Abort to call SIGQUIT (backtrace and coredump) instead of MPI_Abort, also a runtime configurable option.

The note they gave me today is that users would probably want to set this in code. This is something we can discuss as it probably would exist as a flag for the VM.

theurich:
After taking a good look at this PR, I am now wondering if the `ESMF_RUNTIME_LOGMSGABORT` should apply to all the logs a user might open during the execution of the application, or only apply the ESMF default log. My assumption had always been that this would only apply to the ESMF default log, however, the current implementation under this PR would have it apply to any log, even those explicitly opened by the user.  I think this behavior might be unexpected.  The code could easily be changed to only apply `ESMF_RUNTIME_LOGMSGABORT` to the ESMF default log, i.e. the one opened by calling `ESMF_LogInitialize()`,  and not generally to `ESMF_LogOpen()`.

theurich (changes requested):
Generally this looks good to me. But I do have the higher level question in the other comment we should discuss. That might affect some implementation details.

Also documentation for the new `ESMF_RUNTIME_` environment variables needs to be added to show up under the `ESMF_Initialize()` API doc.

danrosen25:
@theurich 
I moved modifications of `alog%logmsgAbort` into `ESMF_LogInitialize`, which is the subroutine that opens `ESMF_LogDefault`. I am looking for a good example for testing this function, it will need to support the application crashing (SIGABRT) and either be set through an environment variable or a config file in ESMF_Initialize. Do you have any pointers?

danrosen25:
@theurich Also added documentation.

danrosen25:
@theurich I believe this is ready for your review again.

src/Infrastructure/Util/src/ESMF_LogErr.F90
```diff
@@ -1730,11 +1730,80 @@ subroutine ESMF_LogInitialize(filename, logappendflag, logkindflag, rc)
...
+    if (msgAbortCnt > 6) then
+      write (ESMF_UtilIOStderr,*) ESMF_METHOD,  &
+          ": msgAbortLst overflow - ESMF_RUNTIME_LOGMSGABORT"
+      call ESMF_UtilIOUnitFlush(ESMF_UtilIOStderr, rc=rc)
+      if (present(rc)) rc=ESMF_RC_BUFFER_SHORT
+      return
+    elseif (msgAbortCnt > 0) then
+      if (associated (alog%logmsgAbort)) deallocate (alog%logmsgAbort)
+      allocate(alog%logmsgAbort(msgAbortCnt))
+      alog%logmsgAbort=msgAbortLst(1:msgAbortCnt)
```
theurich, line 1803: As per comment to use ESMF_LogSet().
```suggestion
      call ESMF_LogSet(logmsgAbort=msgAbortLst(1:msgAbortCnt), rc=localrc)
      if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
        ESMF_CONTEXT, rcToReturn=rc)) return
```

src/Infrastructure/Util/src/ESMF_LogErr.F90
```diff
@@ -1730,11 +1730,80 @@ subroutine ESMF_LogInitialize(filename, logappendflag, logkindflag, rc)
...
+    if (localrc /= ESMF_SUCCESS) then
+      write (ESMF_UtilIOStderr,*) ESMF_METHOD,  &
+          ": ESMF_LogOpen - ESMF_LogDefault"
+      call ESMF_UtilIOUnitFlush(ESMF_UtilIOStderr, rc=rc)
+      if (present(rc)) rc=localrc
+      return
+    end if
+
+    alog => ESMF_LogTable(ESMF_LogDefault%logTableIndex)
+    call c_ESMC_VMGetEnv("ESMF_RUNTIME_LOGMSGABORT", envRtAbort, localrc)
```
theurich, line 1755: Too bad this file in `Util` is compiled _before_ `ESMF_VMGetEnv()` is available. Using the `c_ESMC_` interface, as you have it, is the ugly but correct solution.

src/Infrastructure/Util/src/ESMF_LogErr.F90
```diff
@@ -1730,11 +1730,80 @@ subroutine ESMF_LogInitialize(filename, logappendflag, logkindflag, rc)
...
     call ESMF_LogOpen(ESMF_LogDefault, filename,  &
-        appendflag=logappendflag, logkindflag=logkindflag, rc=rc)
+        appendflag=logappendflag, logkindflag=logkindflag, rc=localrc)
+    if (localrc /= ESMF_SUCCESS) then
+      write (ESMF_UtilIOStderr,*) ESMF_METHOD,  &
+          ": ESMF_LogOpen - ESMF_LogDefault"
+      call ESMF_UtilIOUnitFlush(ESMF_UtilIOStderr, rc=rc)
+      if (present(rc)) rc=localrc
+      return
+    end if
```
theurich, line 1752: Should use standard error handling via `ESMF_LogFoundError()`. I think it is historical that this file isn't doing that consistently. Let's at least do it for new code that being added to this file.

```suggestion
    if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
        ESMF_CONTEXT, rcToReturn=rc)) return
```

src/Infrastructure/Util/src/ESMF_LogErr.F90
```diff
@@ -1730,11 +1730,80 @@ subroutine ESMF_LogInitialize(filename, logappendflag, logkindflag, rc)
...
+
+    alog => ESMF_LogTable(ESMF_LogDefault%logTableIndex)
+    call c_ESMC_VMGetEnv("ESMF_RUNTIME_LOGMSGABORT", envRtAbort, localrc)
+    if (localrc /= ESMF_SUCCESS) then
+      write (ESMF_UtilIOStderr,*) ESMF_METHOD,  &
+          ": VMGetEnv - ESMF_RUNTIME_LOGMSGABORT"
+      call ESMF_UtilIOUnitFlush(ESMF_UtilIOStderr, rc=rc)
+      if (present(rc)) rc=localrc
+      return
+    end if
```
theurich, line 1762: Again standard error handling is preferred:
```suggestion
    if (ESMF_LogFoundError(localrc, ESMF_ERR_PASSTHRU, &
        ESMF_CONTEXT, rcToReturn=rc)) return
```

src/Infrastructure/Util/src/ESMF_LogErr.F90
```diff
@@ -1730,11 +1730,80 @@ subroutine ESMF_LogInitialize(filename, logappendflag, logkindflag, rc)
...
+    idxRtAbort = index (envRtAbort, "ESMF_LOGMSG_WARNING")
+    if (idxRtAbort > 0) then
+      msgAbortCnt = msgAbortCnt + 1
+      msgAbortLst(msgAbortCnt) = ESMF_LOGMSG_WARNING
+    endif
+    if (msgAbortCnt > 6) then
+      write (ESMF_UtilIOStderr,*) ESMF_METHOD,  &
+          ": msgAbortLst overflow - ESMF_RUNTIME_LOGMSGABORT"
+      call ESMF_UtilIOUnitFlush(ESMF_UtilIOStderr, rc=rc)
+      if (present(rc)) rc=ESMF_RC_BUFFER_SHORT
```
theurich, line 1798: Standard error handling for locally detected error condition:

```suggestion
      call ESMF_LogSetError(ESMF_RC_BUFFER_SHORT, &
        msg="msgAbortLst overflow - ESMF_RUNTIME_LOGMSGABORT", &
        ESMF_CONTEXT, rcToReturn=rc)
```

theurich (changes requested):
I think this looks really good, and I am looking forward to using these additional debugging options in my own work soon.

I made a few suggested changes. Since I did those suggestions just on the GitHub interface, I haven't even tried compiling, let along running with those suggested code changes. Please keep that in mind as you accept them. The suggestions are mostly for overall consistent usage patterns across the library.

danrosen25:
@theurich 
Would you prefer `ESMF_RUNTIME_ABORT_LOGMSG` or `ESMF_RUNTIME_ABORT_ON_LOGMSG`?

theurich:
> @theurich Would you prefer `ESMF_RUNTIME_ABORT_LOGMSG` or `ESMF_RUNTIME_ABORT_ON_LOGMSG`?

The advantage of `ESMF_RUNTIME_ABORT_LOGMSG` is shortness, and I like that. The `ESMF_RUNTIME_ABORT_ON_LOGMSG` is probably a little clearer, but a user probably still has to read the documentation to know exactly what it is doing, so I am not sure how much is gained by the extra length of the variable itself.

I am tagging the team @oehmke @billsacks @uturuncoglu @anntsay. It seems a good thing to have consensus on with different perspectives weighing in.

danrosen25:
@oehmke @billsacks @uturuncoglu @anntsay
If you want to read the documentation for what `ESMF_RUNTIME_ABORT_LOGMSG` / `ESMF_RUNTIME_ABORT_ON_LOGMSG` does then you can see it here
https://github.com/danrosen25/esmf/blob/c201260f732c45d8e2b6add18a938cd6ae12df5f/src/doc/ESMF_debug.tex#L159
but it's currently called `ESMF_RUNTIME_LOGMSGABORT`

Once I get a little more input then I'll proceed. Thanks!

src/doc/ESMF_debug.tex
```diff
@@ -155,3 +155,24 @@
...
 In fact the location of the division by zero is correctly identified by the
 runtime library at line 149 in file {\tt ocn.F90}.
+
+It is possible to force a hard crash within the ESMF library while logging
+ERRORS or WARNINGS to the {\tt PET<nnn>.ESMF\_LogFile}. Doing so can be
+advantageous because it may produces a code dump and/or backtrace at the
+initial point of error without ESMF return code handling. There are two
+settings controlling log message error handling. The first
+{\tt ESMF\_RUNTIME\_LOGMSGABORT} configures the ESMF library to abort during
+specified log message types, such as {\tt ESMF\_LOGMSG\_ERROR}. The second
```
billsacks, line 165: It would help to provide an example of how this can be set. My understanding from looking briefly at the code is that this can be set to a single log message type or a list of multiple types. That wasn't totally clear to me from reading through the documentation and I think a couple of examples would help clarify this.

billsacks:
Regarding the question of `ESMF_RUNTIME_ABORT_ON_LOGMSG` vs `ESMF_RUNTIME_ABORT_LOGMSG`: My understanding is that this can be set to one or more logmsg types. If so, I think the clearest thing would be to label it `ESMF_RUNTIME_ABORT_LOGMSG_TYPES`. I know it's verbose, but to me:
- `ESMF_RUNTIME_ABORT_LOGMSG` suggests a message that will be written to the log upon aborting
- `ESMF_RUNTIME_ABORT_ON_LOGMSG` suggests a logical flag

billsacks (commented):
Thanks a lot for this great feature! From a skim through parts of this, I have a few minor comments on the documentation.

billsacks:
> I think I should also change ESMF_RUNTIME_ABORT to ESMF_RUNTIME_ABORT_ACTION for clarity. Okay?

I hadn't thought about that one, but yes: I agree. I like the symmetry of having the last word in the name tell you what this is: this is where you provide an action, and the other one is where you provide the logmsg types.

theurich:
> > I think I should also change ESMF_RUNTIME_ABORT to ESMF_RUNTIME_ABORT_ACTION for clarity. Okay?
> 
> I hadn't thought about that one, but yes: I agree. I like the symmetry of having the last word in the name tell you what this is: this is where you provide an action, and the other one is where you provide the logmsg types.

That makes sense to me, too!

oehmke:
Me too!

src/addon/ESMX/README.md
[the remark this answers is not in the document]
danrosen25: Changed to ESMF_RUNTIME_ABORT_LOGMSG_TYPES

billsacks (approved):
This looks great to me now: between the new variable names and the updated documentation, I find this more clear. Thank you @danrosen25 !

PR #363, opened 2025-03-12
Use correct NetCDF types when writing attributes

billsacks:
This resolves esmf-org/esmf-support#507 (reported by @uturuncoglu) - that NetCDF attributes were being written with the wrong types.

To do this, I extracted some existing NetCDF utility functions (which were private to the class IO_NetCDF) into a new utility module.

I have also added some error checking of the return value of these functions in a couple of places where, previously, these errors were unhandled.


billsacks:
@uturuncoglu - I have tested that this fixes the issue in the little test program that you helped me create, so I expect this will resolve your issue, but it would be great if you're able to test with this branch and verify that.

@oehmke I'd welcome a quick review of this to verify that my creation of these standalone utility functions follows ESMF best practices, e.g., with the use of namespaces. (I found a few examples in the ESMF code, but it didn't feel like things were being done consistently, so I made my best attempt at what felt like good practices.)

I still want to:
- Add some documentation of the extracted functions
- Add some unit testing of the extracted functions

src/Infrastructure/IO/include/ESMCI_IO_NetCDF_Utils.h
```diff
@@ -0,0 +1,43 @@
...
+// Some definitions needed to allow things to work without either ESMF_PNETCDF or
+// ESMF_NETCDF (to avoid needing an overly-complex set of ifdefs elsewhere).
+typedef int nc_type;
+#define NC_NOERR 0
+#endif
+
+#define NC_UNSPECIFIED ((nc_type)0)
+
+namespace ESMCI{
+   namespace NetCDFUtils{
```
billsacks, line 36: Note that I put these functions in a namespace nested inside the ESMCI namespace. Is this a reasonable approach?

src/Infrastructure/IO/src/ESMCI_IO_NetCDF.C
```diff
@@ -743,7 +640,13 @@ void IO_NetCDF::destruct(void) {
...
     }
 
-    ESMC_TypeKind_Flag  arrayType = ncToEsmcType(nctype);
+    ESMC_TypeKind_Flag  arrayType = NetCDFUtils::ncToEsmcType(nctype);
+    if (arrayType == ESMF_NOKIND) {
+      string errstr = string(": problem converting NetCDF type to ESMF type");
+      ESMC_LogDefault.Write(errstr, ESMC_LOGMSG_ERROR, ESMC_CONTEXT);
+      *rc = ESMF_FAILURE;
+      return thisArray;
+    }
```
billsacks, line 649: Note the error handling that I've added here. The original function didn't have an rc return value, and I kept that interface – but I could change this to have an explict rc if you feel that would be better. (I'm not sure what the accepted approach is for low-level internal utility functions like this.)

src/Infrastructure/IO/src/ESMCI_IO_NetCDF.C
```diff
@@ -980,7 +883,12 @@ void IO_NetCDF::destruct(void) {
 
     ESMC_TypeKind_Flag  esmcType = thisArray->getTypekind();
 //printf("ESMC Type: %d\n", esmcType);
-    nc_type            ncType = esmcToNcType(esmcType);
+    nc_type            ncType = NetCDFUtils::esmcToNcType(esmcType);
+    if (ncType == NC_UNSPECIFIED) {
+      string errstr = string(": problem converting ESMF type to NetCDF type");
+      ESMC_LogDefault.MsgFoundError(ESMF_FAILURE, errstr, ESMC_CONTEXT, &localrc);
+      return localrc;
+    }
```
billsacks, line 891: As above: Note the error handling that I've added here. The original function didn't have an rc return value, and I kept that interface – but I could change this to have an explict rc if you feel that would be better. (I'm not sure what the accepted approach is for low-level internal utility functions like this. Also, for this function, note that, in the new usage in the PIO handler, I deliberately don't check the return value because there may be cases where there's an unhandled type but that's okay.)

src/Infrastructure/IO/src/ESMCI_IO_NetCDF.C
```diff
@@ -1085,7 +993,9 @@ void IO_NetCDF::destruct(void) {
       break;
 
     default:
-      break;
+      string errstr = string(": unhandled NetCDF type");
+      ESMC_LogDefault.MsgFoundError(ESMF_FAILURE, errstr, ESMC_CONTEXT, &localrc);
+      return localrc;
```
billsacks, line 998: Note some more error handling here. I think it's correct to add this error handling instead of the previous behavior of quietly doing nothing in this case: I can't think of a reason why you'd *want* it to quietly do nothing here.

src/Infrastructure/IO/src/ESMCI_IO_NetCDF_Utils.C
```diff
@@ -0,0 +1,138 @@
...
+
+//-------------------------------------------------------------------------
+#undef ESMC_METHOD
+#define ESMC_METHOD "ESMCI::NetCDFUtils::ncToEsmcType"
+ESMC_TypeKind_Flag ESMCI::NetCDFUtils::ncToEsmcType(nc_type ncTypeVal)
+{
+
+   ESMC_TypeKind_Flag esmcTypeVal = ESMF_NOKIND;
+
+#if defined(ESMF_NETCDF) || defined(ESMF_PNETCDF)
```
billsacks, line 57: I considered a few ways to do the conditional compilation here and in esmcToNcType. Based on a quick code analysis, I think we need these functions to be compiled even if we're not using netcdf or pnetcdf, but the return value is unimportant in those cases. I don't love putting essentially the whole body of the function inside an `#if`, but this seemed like the best way to do it - though I'm open to other suggestions.

(As a side note: I'd love if we could just say that netcdf is required, or have some intermediate layer between netcdf and esmf that we use from esmf to avoid a direct dependency, so we could avoid having `#if`s like this in the code, because I find they make it much harder to read, write and modify code.)

src/Infrastructure/IO/include/ESMCI_IO_NetCDF_Utils.h
oehmke: Aside from ESMCI, I usually don't use namespaces. If there's a good reason to do, for example,  because you're worried about a clash with similar names in other parts of ESMCI, then it seems fine, otherwise it seems like overkill. Just out of curiosity, did you find nested namespaces elsewhere in ESMF?   All that being said, the C++ side has always been a little more loose, so if you feel like you really want a nested namespace, go ahead... :-)

src/Infrastructure/IO/src/ESMCI_IO_NetCDF.C
oehmke: That's great that you added error handling! Usually, however, we use ESMC_LogDefault.MsgFoundError() and then for the first entry we use one of the error codes from: https://earthsystemmodeling.org/docs/nightly/develop/ESMF_refdoc/node9.html#SECTION09030000000000000000
You can see the associated messages here: <ESMF_DIR>/src/Infrastructure/LogErr/include/ESMCI_ErrMsgs.C

The codes/messages don't cover everything, but just pick something close or you can add a new one if it seems necessary.  There is a generic netcdf one:  ESMC_RC_NETCDF_ERROR. 

I"m not sure why the older code is just setting things to ESMF_FAILURE? 

I'm not sure why

src/Infrastructure/IO/src/ESMCI_IO_NetCDF.C
oehmke: See above.

src/Infrastructure/IO/src/ESMCI_IO_NetCDF.C
oehmke: I completely agree!

oehmke (commented):
Hi Bill,  Overall this looks really good. I'm glad that you went through and added more error checking and reorganized the code to have a better structure. The only thing that I would say to change is to return a specific error code instead of ESMF_FAILURE. Thanks for taking this on.

src/Infrastructure/IO/src/ESMCI_IO_NetCDF.C
billsacks: > That's great that you added error handling! Usually, however, we use ESMC_LogDefault.MsgFoundError() and then for the first entry we use one of the error codes from: https://earthsystemmodeling.org/docs/nightly/develop/ESMF_refdoc/node9.html#SECTION09030000000000000000 You can see the associated messages here: <ESMF_DIR>/src/Infrastructure/LogErr/include/ESMCI_ErrMsgs.C
> 
> The codes/messages don't cover everything, but just pick something close or you can add a new one if it seems necessary. There is a generic netcdf one: ESMC_RC_NETCDF_ERROR.
> 
> I"m not sure why the older code is just setting things to ESMF_FAILURE?

@oehmke - thanks a lot for this guidance. I followed the examples of the surrounding code in this function, but it sounds like they aren't following ESMF best practices. I'll change these to something like:

```C++
ESMC_LogDefault.MsgFoundError(ESMC_RC_NETCDF_ERROR, errstr, ESMC_CONTEXT, &localrc)
```

(but I'll see if I can find a more appropriate error code).

uturuncoglu:
@billsacks I could able to test your branch with the UFS WM head of develop. It seems it is working as expected and fixes the issue. To test it I run one of the regression test of UFS WM with tile output and compare the netcdf files with ESMF 8.6.0. The test that I am using is `datm_cdeps_lnd_gswp3_intel` (CDEPS Data Atmosphere coupled with NoahMP).

**Run with ESMF 8.6.0:**
```
$ ncdump -c /work2/noaa/stmp/tufuk/stmp/tufuk/FV3_RT/rt_2825166/datm_cdeps_lnd_gswp3_intel/ufs.cpld.lnd.out.2000-01-02-00000.tile1.nc | grep mask

	int mask(time, grid_yt, grid_xt) ;
		mask:long_name = "land-sea mask" ;
		mask:missing_value = -999s ;
		mask:units = "1" ;
```

**Test with your branch:**
```
$ ncdump -c /work2/noaa/stmp/tufuk/stmp/tufuk/FV3_RT/rt_2839393/datm_cdeps_lnd_gswp3_intel/ufs.cpld.lnd.out.2000-01-02-00000.tile1.nc | grep mask

	int mask(time, grid_yt, grid_xt) ;
		mask:long_name = "land-sea mask" ;
		mask:missing_value = -999 ;
		mask:units = "1" ;
```

As you can see, the missing value is short type in the old version but set as integer correctly in your branch. I also check other variables and they seem fine.

uturuncoglu (approved):
It looks good to me.

billsacks:
@oehmke - thank you very much for your review; and @uturuncoglu - thank you very much for testing this.

@oehmke - I have made changes to address the good points you raised in your review. Please feel free to re-review or just give me the go-ahead to merge it whenever you get a chance.

oehmke:
I looked over your commits, and they look good to me. Feel free to go ahead and merge. Thanks!

PR #368, opened 2025-03-26
Give a better error message for ESMF_Initialize after ESMF_Finalize

billsacks:
The previous error message in this case was more cryptic:

```
ESMF_LogWrite(): ESMF_Log not open -- cannot ESMF_LogWrite().  Log message = Object being used after deletion  - Bad Object
```

The new error message is:

```
ESMF_FrameworkInternalInit: Cannot reinitialize ESMF after it has been finalized
```

This case seems hard (or maybe impossible) to test in the unit testing framework, but I have tested it with the code here:

https://github.com/billsacks/esmfprojects-multiple_esmf_initializes


billsacks:
@oehmke - I'd welcome a quick review whenever you get a chance

anntsay:
Team reviewed during core team meeting and approved this PR.

PR #369, opened 2025-03-28
Make the ESMX CMake API directly usable for (complex) projects that leverage CMake

theurich:
This PR makes a few modifications so that the top-level ESMX directory can directly be added to user level `CMakeLists.txt` files via the CMake `add_subdirectory()` function. This allows projects that already leverage CMake to easily add ESMX to the existing build system. It also simplifies more complex cases where components might have complex package dependencies, and where it is more convenient to define these CMake targets explicitly in a `CMakeLists.txt` file before using the target via `esmxBuild.yaml` in ESMX.

The documentation in `README.md` has been updated to discuss the new features.

The ESMX examples under the NUOPC app protos repo (branch feature/esmx-cmake-api) have been updated to adjust to the changes, and to test and demonstrate the new features.


src/addon/ESMX/Driver/CMakeLists.txt
```diff
@@ -415,53 +398,79 @@ foreach(CMP IN ITEMS ${COMPS})
...
+    endif()
+  elseif(${CMP_BUILD_TYPE} STREQUAL "cmake")
+    message(STATUS "ESMX Found CMake subdirectory: ${CMP_SOURCE_DIR}")
+    is_subdirectory(${CMP_SOURCE_DIR} ${CMAKE_CURRENT_SOURCE_DIR} IS_SUB)
+    if(IS_SUB)
+      # component source dir under current source dir
+      add_subdirectory(${CMP_SOURCE_DIR})
+    else()
+      # component source dir NOT under current source dir
+      add_subdirectory(${CMP_SOURCE_DIR} ${CMAKE_BINARY_DIR}/${CMP})
```
danrosen25, line 415: Is it possible to always use this line whether or not the component is a subdirectory?

src/addon/ESMX/Driver/CMakeLists.txt
```diff
@@ -318,45 +339,7 @@ foreach(CMP IN ITEMS ${COMPS})
 
   # build component with cmake
   if(${CMP_BUILD_TYPE} STREQUAL "cmake")
-    message(STATUS "ESMX building (CMAKE) - ${CMP} - ${CMP_SOURCE_DIR}")
```
danrosen25, line 321: It looks like you're gutting the ability to build a CMake project externally. This means that all parent variables are set for child components. This may be undesirable. You may want to consider having both functionalities or both `add_subdirectory` and `ExternalProject_Add`

danrosen25 (commented):
Let's discuss the changes before I approve. I listed some comments that we can address in a meeting. I'm available most of this week.

src/addon/ESMX/CMakeLists.txt
```diff
@@ -1,34 +1,42 @@
...
+  set(CMAKE_CXX_COMPILER            "${ESMF_CXXCOMPILER}")
+endif()
+if(NOT DEFINED CMAKE_C_COMPILER)
+  set(CMAKE_C_COMPILER              "${ESMF_CCOMPILER}")
+endif()
 
 # Project
 project(ESMX
-        VERSION 1.0.0
+        VERSION 8.9.0
```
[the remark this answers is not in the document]
theurich: Good point, there is probably a better way to construct the ESMF release version and use here via variable.  The main point was to align all the `project()` versions under ESMX with the ESMF version under which it is released.

src/addon/ESMX/Comps/ESMX_Data/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.22)
 
-list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
```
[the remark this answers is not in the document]
theurich: Yes, if you _copy_ the `ESMX_Data` directory somewhere else now, e.g. to customize it, you will have to edit the `CMakeLists.txt` file to point to your project's CMake module location, which presumably already contains `FindESMF.cmake`.  I think that is actually more preferable than bringing in yet another (potentially different version of) `FindESMF.cmake`.

src/addon/ESMX/Driver/CMakeLists.txt
```diff
@@ -16,14 +16,22 @@ function(esmx_check_ret RETVAL MESSAGE)
   endif()
 endfunction()
 
-# unset user ESMF variables stored in cache
```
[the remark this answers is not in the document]
theurich: Wow, I have no recollection of even removing this. ... but I see the commit where I did, when I added `function(is_subdirectory ...)`. I think it was by accident that I delete the code you are referring to. I will bring it back in on my local repo and re-run the tests to make sure it's all good. If so I will push to bring this section back. Sorry about that!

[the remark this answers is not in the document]
theurich: Testing with this code block restored is clean for the prototypes. The restoring change is pushed.

src/addon/ESMX/Driver/CMakeLists.txt
theurich: Yes, that is the key CMake feature that allows us to integrate `ESMX` from its installation location into any other project's CMake build system. The only requirement (from CMake) is that when the provided source directory is not a subdir, then you must provide the binary directory of where CMake is supposed to put the build artifacts - since it doesn't know where to integrate it automatically. It's a great feature of CMake.

danrosen25: I mean to clean this up to always use `add_subdirectory(${CMP_SOURCE_DIR} ${CMAKE_BINARY_DIR}/${CMP})` and remove `add_subdirectory(${CMP_SOURCE_DIR})`

src/addon/ESMX/Driver/CMakeLists.txt
theurich: I see your point about the parent variables being set for the child component. It seems though that for most cases that is actually the desired behavior. You do want a consistent build of the final application, and having consistent parent variables available helps with that. Another important behavior the changed version achieves is that the child CMake build now integrates naturally into the parent configure/build/install sequence. What I mean by that is that it's unexpected (for most CMake users) to see a full configure/build/install of a subcomponent during the parent configure step.

Another real practical issue I ran into with the external CMake build approach was bringing up all of the package dependencies (of a complex component) to the parent level during linking. The configuration *.cmake file let the parent know about those packages, but the parent essentially had to use `find_package()` explicitly on all of them to make them available during linking so the child component would successfully link. In the more integrated approach CMake takes care of all of that automatically. If we implement an `ExternalProject_add` (which I do agree might be nice to have!) I think we need to figure out whether there is a better way to promote package dependencies up the hierarchy.

src/addon/ESMX/Driver/CMakeLists.txt
theurich: That would probably work, but seems better to just let CMake do its thing for real subdirs.

theurich: After further discussion on the core call today, it was decided that simplifying the logic, and eliminating the need for the `is_subdirectory`  function is large enough a benefit to make the change. The change has been pushed.

src/addon/ESMX/Driver/CMakeLists.txt
theurich: We decided to add a new component `build_type`: `cmake.external`. A component that sets `cmake.external` will be built by ESMX using the explicit `cmake` calls as previously build type `cmake` was doing by default. Changes have been pushed.

theurich:
@danrosen25 I think this ready for final review. I made the two changes as per the discussion earlier today.

theurich:
@danrosen25 just a heads up that I am planning on merging this into develop in about 30 mins so it will be picked up by nightly testing. I made the changes we talked about, so I don't think it strictly needs another review.

danrosen25 (approved):

PR #372, opened 2025-04-02
Fix for spack CI/CD testing

uturuncoglu:
This PR  fixes Spack installation GitHub action. It was failing because of changed some conventions like compilers.yaml -> packages.yaml to keep list of available compilers in the spack side.

PR #374, opened 2025-04-02
Consistent CMake 3.5.2 version requirement under PIO

theurich:
Currently the [ESMF All Tests](https://github.com/esmf-org/esmf/actions/workflows/esmf-all-tests.yml) action is failing because of the GH runners using CMake 4.0.0 as described by @danrosen25 in https://github.com/esmf-org/esmf/pull/372#issuecomment-2773603691

This PR provides a quick fix for this by moving the CMake version requirement for anything under `src/Infrastructure/IO/PIO/ParallelIO/` to consistently `cmake_minimum_required (VERSION 3.5.2)`. This is expected to not trigger an issue with CMake 4.0.0.

### DIFF
--- a/src/Infrastructure/IO/PIO/ParallelIO/src/clib/CMakeLists.txt
+++ b/src/Infrastructure/IO/PIO/ParallelIO/src/clib/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 2.8.12)
+cmake_minimum_required (VERSION 3.5.2)
 include (CheckFunctionExists)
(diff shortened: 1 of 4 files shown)


danrosen25:
If All Tests pass then I'm ready to go. I've confirmed that the tests on Ubuntu are running CMake 4.0.0. It seems like as GitHub progresses through deploying the Ubuntu images they are all updating to CMake 4.0.0.

billsacks (approved):
Thanks for this fix!

danrosen25 (approved):
Works like a charm.

PR #375, opened 2025-04-03
Cleanup README badges

danrosen25:
* rename workflows
* add NUOPC App Prototype badge
* move doc badges to end


theurich (approved):
I like it!

PR #348, opened 2025-01-29
Fixes esmf-org/esmf#136, fix cygwin build

harmenwierenga:
At Deltares we rely on a Cygwin build of ESMF. There was an attempt at updating ESMF two few years ago, see https://github.com/esmf-org/esmf/issues/136. I have now managed to create a working build of ESMF 8.8.0, and it required me to make a few changes to build everything (including PIO) for Windows. 

I installed the following packages in cygwin (not 100% sure that this list is exhaustive, since some other packages were already installed):
- gcc-core 12.4.0-3
- gcc-fortran 12.4.0-3
- gcc-g++ 12.4.0-3
- make 4.4.1-2
- libnetcdf-devel 4.9.2-2
- libnetcdf-fortran-devel 4.5.4-1
- zlib-devel 1.3.1-1
- libhwloc-devel 2.10.0-1
- libevent-devel 2.1.12-1
- cmake 3.28.3-1

```
set -o igncr
export ESMF_DIR=/cygdrive/c/<ESMF_CHECKOUT_DIR>
export ESMF_COMM=mpiuni
export ESMF_OPTLEVEL=2
export ESMF_NETCDF=split
export ESMF_NETCDF_INCLUDE=/usr/include
export ESMF_NETCDF_LIBPATH=/usr/lib
export ESMF_INSTALL_PREFIX=install
export ESMF_CXXSTD=sysdefault # some C++ libraries require gcc posix extensions, specifically the sigaction definition
export ESMF_RANLIB=true # ranlib does not work during the install stage, use this no-op
```

The issues that I encountered:

1. Some POSIX headers are used, and the default C++ standard `-std=c++11` turns off GNU extensions of the compiler. I had to `export ESMF_CXXSTD=sysdefault` to ensure that the standard used is `-std=gnu++17` so that GNU extensions were turned on. Going to C++17 did not give any issues.
2. In the PIO CMake (src/Infrastructure/IO/PIO/ParallelIO/src/clib/CMakeLists.txt), there is a check that long long and size_t are the same size. On cygwin, these sizes cannot be retrieved, which makes them both empty strings. This causes the check to fail, because empty strings cannot be parsed as numbers so `${SIZEOF_SIZE_T} EQUAL ${SIZEOF_LONG_LONG}` evaluates to false in CMake. I have added a small fix for this that does nothing if either returns an empty string.
[...]

PR #376, opened 2025-04-04
Reorder PIO to occur before NetCDF in common.mk

billsacks:
This is needed with the linker on cygwin since PIO uses NetCDF, so the link flags need to be ordered -lpioc -lnetcdf. (It's possible that other linkers need this also, though we haven't seen the need for it on our test systems.)

This replaces #348 , based on decided-upon solution from this week's ESMF core team meeting.

FYI @theurich @danrosen25 @harmenwierenga 

@harmenwierenga - I took your commit from your earlier branch that reorders the PIO and NETCDF sections in `common.mk`, because after further discussion, we saw rationale for that approach.

@theurich and @danrosen25 - let me know if you'd like to look this over before I merge it.

### DIFF
--- a/build/common.mk
+++ b/build/common.mk
@@ -1460,6 +1460,53 @@ endif
 endif
 
 #-------------------------------------------------------------------------------
+# PIO
+#-------------------------------------------------------------------------------
+ifneq ($(origin ESMF_PIO), environment)
+ifndef ESMF_PIO
+export ESMF_PIO = $(ESMF_PIODEFAULT)
+endif
+
+ifeq ($(ESMF_PIO),internal)
+ifndef ESMF_NETCDF
+# PIO, starting with version 2, depends on NetCDF. Defaulting to internal needs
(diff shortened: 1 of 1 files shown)


DWesl:
Since this is based on environment variables rather than attempted compilation, this should work fine

> It's possible that other linkers need this also, though we haven't seen the need for it on our test systems

Windows is the only other system I've noticed with that link-order requirement, though it seems to be weaker: there's another project that probably had Windows CI that apparently handled `-lblas -llapack` just fine before I reversed the order so it would compile on Cygwin.

I'm told there's a `-no-undefined` linker flag that should emulate some of the Windows linker semantics, but I don't know which compiler supports it nor how much of the Windows semantics it implements.

harmenwierenga:
@billsacks Thank you for this! I have not had time to look into testing it unfortunately. PIO accepted my changes to their repo a while ago, would it be feasible to update the version of PIO that is used by ESMF?

harmenwierenga:
Nice, thank you!

PR #377, opened 2025-04-08
Tone down the restrictions for the clock object in *CompCreate() API doc

theurich:
Resolves https://github.com/esmf-org/esmf-support/issues/503.


theurich:
@oehmke this minor doc change might be a quick one to review and move forward. Thanks!

oehmke (approved):
That sounds much clearer to me.  I like how you described the intended usage, but then explained why the user can use the clock as they'd like as long as they are consistent. Thanks!

PR #390, opened 2025-04-29
Add workflow for testing MAPL

danrosen25:
This workflow will run once a week to maintain cache. It can be run manually using workflow dispatch.

Options
- ESMF Branch/Tag (esmf-vers default: 'develop')
- MAPL Repository (mapl-repo default: 'GEOS-ESM/MAPL')
- MAPL Branch/Tag (mapl-vers default: 'release/MAPL-v3')
- Enable Artifacts (enbl-artf default: false, artifacts are always created if workflow fails.)

### DIFF
--- a/.github/workflows/mapl-tests.yml
+++ b/.github/workflows/mapl-tests.yml
@@ -1,0 +1,280 @@
+name: MAPL Test
+
+on:
+  schedule:
+    - cron: "15 9 * * 3"
+  workflow_dispatch:
+    inputs:
+      mapl-repo:
+        description: 'MAPL Repository'
+        required: true
+        default: 'GEOS-ESM/MAPL'
+      mapl-vers:
+        description: 'MAPL Branch/Tag'
+        required: true
+        default: 'develop'
+      enbl-artf:
+        description: 'Enable Artifacts'
+        required: true
+        type: boolean
+        default: false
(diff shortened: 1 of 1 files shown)


oehmke:
We talked about this on the NASA call. They said that it would be best to do: release/MAPL-v3  as the default to run periodically.

danrosen25:
Hi @oehmke,
I tried multiple variants of OpenMPI v4.1.6 and v5.0.7 with the `release/MAPL-v3` and I'm seeing test failures. I also tried setting `OMPI_MCA_osc=^pt2pt` for both versions tests still failed. I can't tell why this setup fails but the GEOS MAPL test is working. I have changed the default ref to `develop` using OpenMPI v4.1.6.

oehmke:
Just so I make sure that understand where we are: 
+ The MAPL with an ESMF tests works in our repository with GEOS MAPL develop
+ For some reason you are getting errors with release/MAPL-v3

Is that right? 

If this is working with develop for now and you need to switch back to something else for a while, that’s fine. At some point it would be good to figure this out, but I understand that you also have other priorities.

oehmke (approved):
I don't do a lot of stuff with actions, but from what I can tell it looks good to me. Thanks for putting this together!

PR #393, opened 2025-04-29
Update ESMF_FieldBundle.cppF90

tclune:
The SetGeom() procedure exists, but it was not added to the generic name.


theurich:
@oehmke do you approve?

oehmke (approved):
Thanks for the fix! Looks good to me.

Issue #254, opened 2024-05-30
Error building with gfortran on 32 bit ABI machine


danrosen25:
### Discussed in https://github.com/orgs/esmf-org/discussions/251

<div type='discussions-op-text'>

<sup>Originally posted by **GrigorisLionis** May 28, 2024</sup>

### ESMF Version

v.8.6.1.

### Issue

I am trying to build latest stable version from source (a.k.a. 8.6.1.)
I have encountered a problem compiling the sources, due to a type mismatch error
The output of the error is as following:
gfortran -c -fallow-argument-mismatch -fallow-invalid-boz -O -fPIC  -pthread -ffree-line-length-none  -fopenmp -I/home/user/esmf/esmf-8.6.1/src/Infrastructure/Util/src -I/home/user/esmf/esmf-8.6.1/build_config/Linux.gfortran.default -I/home/user/esmf/esmf-8.6.1/src/Infrastructure -I/home/user/esmf/esmf-8.6.1/src/Superstructure -I/home/user/esmf/esmf-8.6.1/mod/modO/Linux.gfortran.32.mpiuni.default -I/home/user/esmf/esmf-8.6.1/src/include   -DESMF_NO_INTEGER_1_BYTE -DESMF_NO_INTEGER_2_BYTE -DESMF_MPIUNI -I/home/user/esmf/esmf-8.6.1/src/Infrastructure/stubs/mpiuni -DESMF_LOWERCASE_SINGLEUNDERSCORE -DESMF_MOAB=1 -DESMF_LAPACK=1 -DESMF_LAPACK_INTERNAL=1 -DESMF_NO_ACC_SOFTWARE_STACK=1 -DESMF_YAMLCPP=1 -DESMF_YAML=1 -DESMF_NO_OPENACC -DESMF_BOPT_O -DESMF_TESTCOMPTUNNEL -DS32=1 -DESMF_OS_Linux=1 -DESMF_COMM=mpiuni -DESMF_DIR=/home/user/esmf/esmf-8.6.1 /home/user/esmf/esmf-8.6.1/src/Infrastructure/Util/src/ESMF_FortranWordsize.F90 -o /home/user/esmf/esmf-8.6.1/obj/objO/Linux.gfortran.32.mpiuni.default/ESMF_FortranWordsize.o
/home/user/esmf/esmf-8.6.1/src/Infrastructure/Util/src/ESMF_FortranWordsize.F90:425:77:

[...]

PR #394, opened 2025-04-29
Fix ESMF_FortranWordsize for 32-bit systems

Dave-Allured:
Pointer size is hard coded as `integer*8` in one place only, in the whole ESMF code base.  This should probably be `integer(C_SIZE_T)` to conform with incoming subroutine calls.  Please confirm, I am not very familiar with ISO_C_MODULE issues.

This may fix an error on 32-bit PPC.  A MacPorts developer is testing ESMF on 32-bit PPC.  Downstream issue has error messages and other details:

https://trac.macports.org/ticket/72424

### DIFF
--- a/src/Infrastructure/Util/src/ESMF_FortranWordsize.cppF90
+++ b/src/Infrastructure/Util/src/ESMF_FortranWordsize.cppF90
@@ -73,7 +73,7 @@ module ESMF_FortranWordsizeMod
   interface
 
     subroutine esmf_pointerdifference(n, s1, s2, len)
-      integer*8 :: n
+      integer(C_SIZE_T) :: n
       type(*) :: s1
       type(*) :: s2
       integer :: len
--- a/src/Infrastructure/Util/src/ESMF_FortranWordsize.cppF90
+++ b/src/Infrastructure/Util/src/ESMF_FortranWordsize.cppF90
@@ -73,6 +73,7 @@ module ESMF_FortranWordsizeMod
   interface
 
     subroutine esmf_pointerdifference(n, s1, s2, len)
+      use iso_c_binding
       integer(C_SIZE_T) :: n
       type(*) :: s1
       type(*) :: s2


theurich:
Reminds me of https://github.com/esmf-org/esmf/issues/254. Just cherry-picked the work done there into a new branch: https://github.com/esmf-org/esmf/tree/fix/32-bit-abi

Dave-Allured:
@theurich Yes indeed, it looks like exactly the same issue to me.  I see the switch to `integer(C_SIZE_T)` in that 8.6.2 side branch, but it seems to have been lost in today's primary `develop` version.

If 32-bit mode has really not been tested in 6 years, then this may be only the first of several 32-bit issues.  Let's see.

Dave-Allured:
Oh, new branch, I had overlooked that.   Thanks for doing that, and let's hope for good results.  That MacPorts developer may be interested in more 32-bit testing via MacPorts.

theurich:
Yep, patch/8.6.2 was never officially tagged. There were run-time failures, and without 32-bit test machine it was going to be difficult to hunt down. Yep, please see how well https://github.com/esmf-org/esmf/tree/fix/32-bit-abi does in testing, and we can go from there.

barracuda156:
Thank you, I will try that branch.

> There were run-time failures, and without 32-bit test machine it was going to be difficult to hunt down.

Since this is just about bitness, 32-bit Linux/*BSD should work just as fine, and, I believe, can be emulated via Qemu.
Probably some Linux distros and perhaps at least two *BSDs (OpenBSD and NetBSD) should have pre-built stuff for i386 and ppc, so it may not be too painful to deal with the builds.
Possibly macOS can be emulated as well, via UTM or Qemu, 10.14 was the last version to support building for i386. I am not really sure what MacPorts archives have pre-built for macOS on i386 (for ppc there is exactly nothing), but likely at least compilers and stuff like cmake are available.

Dave-Allured:
I will close this small pull request in deference to the more comprehensive 32-bit development branch https://github.com/esmf-org/esmf/tree/fix/32-bit-abi by @theurich.  I look forward to the successful completion of that branch.  Thank you Gerhard for the effort.

theurich:
> Thank you, I will try that branch.
> 
> > There were run-time failures, and without 32-bit test machine it was going to be difficult to hunt down.
> 
> Since this is just about bitness, 32-bit Linux/*BSD should work just as fine, and, I believe, can be emulated via Qemu. Probably some Linux distros and perhaps at least two *BSDs (OpenBSD and NetBSD) should have pre-built stuff for i386 and ppc, so it may not be too painful to deal with the builds. Possibly macOS can be emulated as well, via UTM or Qemu, 10.14 was the last version to support building for i386. I am not really sure what MacPorts archives have pre-built for macOS on i386 (for ppc there is exactly nothing), but likely at least compilers and stuff like cmake are available.

@danrosen25 Are there any options on GH that would allow us to test with 32-bit Linux and/or Darwin? Does above info from @barracuda156 help?

danrosen25:
@theurich GitHub only provides x64 and arm64 runners https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories. I'd have to do some research on emulating 32-bit using qemu in a Docker container. Is this something you'd like because we need to open a new issue?

barracuda156:
Sorry, missed this one. Yes, for Linux (should be any arch supported by Qemu), no for Darwin, AFAIK.

PR #397, opened 2025-04-30
More control over Hierarchy Connectors via new NUOPC attribute

theurich:
This PR implements the new NUOPC attribute, `HierarchyConnectors`, which can be set on Driver, Model, and Mediator components. This attribute provides more control over which child components are connected through the component hierarchy.

A need for this increased level of control over hierarchy connectors was discovered in the NRL project where the NEPTUNE atmosphere component is being integrated into the Navy-ESPC coupled systems. The NEPTUNE component uses NUOPC to implement asynchronous output, where each output stream is instantiated via its separate NUOPC_Model component. Without the `HierarchyConnectors = manual` option, hierarchy connectors (between parent and child components) are created for _all_ of the child components. This is inefficient, and in some cases leads to conflicts (e.g. multiple output stream components have the same fields in the import state, which then are mirrored into the parent import state, leading to duplicates, triggering NUOPC ERROR condition).

The `HierarchyConnectors = auto` is the default, which is backward compatible with previous behavior (e.g. `HierarchyConnectors` unset). With `HierarchyConnectors = manual`, the user has full control over which hierarchy connectors are instantiated.


src/addon/NUOPC/src/NUOPC_Driver.F90
```diff
@@ -994,113 +994,120 @@ recursive subroutine InitializeIPDv02p1(driver, importState, exportState, &
...
-                ESMF_StateIsCreated(is%wrap%modelIS(i), rc=rc)
-              if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, &
-                line=__LINE__, file=trim(name)//":"//FILENAME)) return
-            else
-              ! j==0
-              needConnector = needConnector.and. &
-                ESMF_StateIsCreated(is%wrap%modelES(j), rc=rc)
-              if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, &
-                line=__LINE__, file=trim(name)//":"//FILENAME)) return
+          needConnector=.not.needConnector  ! default to do connection
```
danrosen25, line 1007: This is really hard to follow - If a value is set for HierarchyProtocol then needConnector is false (because it's flipped). And if needConnector is false then condition is satisfied. Can you add a second boolean such as `hrchProtSet` then `needConnector=.not.hrchProtSet`.

Then change the next line to `if (hrchProtSet) then`

src/addon/NUOPC/src/NUOPC_Driver.F90
```diff
@@ -994,113 +994,120 @@ recursive subroutine InitializeIPDv02p1(driver, importState, exportState, &
...
+                call NUOPC_DriverGetComp(driver, srcCompLabel, dstCompLabel, &
+                  comp=connector, petList=petList, relaxedflag=.true., rc=rc)
+                if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, &
+                  line=__LINE__, file=trim(name)//":"//FILENAME)) return  ! bail
+                ! automatically created connectors inherit Verbosity from parent
+                call NUOPC_CompAttributeGet(driver, name="Verbosity", &
+                  value=valueString, rc=rc)
+                if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, &
+                  line=__LINE__, file=trim(name)//":"//FILENAME)) return  ! bail
+                if (trim(valueString)=="max" .or. trim(valueString)=="high" .or. &
```
danrosen25, line 1059: Why not all non-blank values?

theurich: Automatically created connectors inherit from the parent component (driver) that creates them. The meaning of Verbosity bits for Driver and Connector  agree for the lower 8 bits, therefore only inherit those, if this is a bit pattern. for presets, they are interpreted according to the NUOPC component kind, so presets can just be inherited directly. Does that make sense?

src/addon/NUOPC/src/NUOPC_Driver.F90
theurich: Yep, I too had to look at the handling of `needConnector` for a bit before I understood it again when I came back to this code for this PR. However, I liked it again the way it was a after I understood it again.... but I also see your point.... so I just simplified it with an extra logical in 9eaf4ff.

src/addon/NUOPC/src/NUOPC_Driver.F90
danrosen25: It would make more sense if the connector contained metadata saying it was automatically created and if it was automatically created then it always inherits the lower 8 bits, instead of checking the value for "max","high", etc

theurich: "max", "high", "low", and "off" are inherited verbatim. Any of these might define higher bits than the lower 8. However, the definition is component kind dependent, and thus must be done verbatim. If not a string, but an explicit bit pattern is found in the parent Verbosity setting, then the handling for inheritance is done as bit pattern, and only lower 8 bits are inherited. So the inheritance handling depends on the parents' Verbosity format. Same applies to Profiling.

danrosen25: Well, that's not exactly what I meant but okay. The setup is a little confusing and therefore it will be difficult to maintain in the future.

danrosen25 (approved):

PR #399, opened 2025-05-05
Unsure expected behavior when using ESMX from CMake based project build

theurich:
This PR removes the setting of custom compiler flags from the top level ESMX `CMakeLists.txt` file. Setting these flags causes unexpected behavior in projects that use CMake and leverage ESMX.

### DIFF
--- a/src/addon/ESMX/CMakeLists.txt
+++ b/src/addon/ESMX/CMakeLists.txt
@@ -28,22 +28,6 @@ include(GNUInstallDirs)
 # Add ESMX driver
 add_subdirectory(Driver)
 
-if(CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
-  set(CMAKE_Fortran_FLAGS_RELEASE "-O2")
-  set(CMAKE_Fortran_FLAGS_RELWITHDEBINFO "-g -fbacktrace -O2")
-  set(CMAKE_Fortran_FLAGS_DEBUG "-g -fbacktrace -O0 -fcheck=all -ffpe-trap=invalid,zero,overflow,underflow")
-elseif(CMAKE_Fortran_COMPILER_ID STREQUAL "Intel")
-  set(CMAKE_Fortran_FLAGS_RELEASE "-O2")
-  set(CMAKE_Fortran_FLAGS_RELWITHDEBINFO "-g -traceback -O2")
-  set(CMAKE_Fortran_FLAGS_DEBUG "-g -traceback -O0 -check all -fpe0 -ftrapuv -init=snan,arrays")
-elseif(CMAKE_Fortran_COMPILER_ID STREQUAL "NVHPC")
-  set(CMAKE_Fortran_FLAGS_RELEASE "-O2 -fast")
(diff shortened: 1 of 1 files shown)


danrosen25 (commented):
Removing these compiler flags will push debugging onto the end-user. I believe that end-users using ESMX don't have the technical expertise to set any compiler flags. I'm guessing the change in unexpected behavior started when `add_subdirectory` was added? Because you don't want subcomponents to pick up compiler flags? The alternative to add_subdirectory would be to use CMake External Projects, which don't pass flags to subcomponents.

theurich:
The problem I am working on resolving comes in when _linking_ the final executable. The individual components pulled in via `add_subdirectory` into a the user controlled top-level `CMakeLists.txt` file are unaffected by ESMX flags (either way), because the `add_subdirectory(${ESMF_ESMXDIR} ${CMAKE_BINARY_DIR}/ESMX)` has to be _after_ the components. However, the final executable is linked with flags provided by ESMX' top-level `CMakeLists.txt` file, and that is unexpected in that case, where the user potentially sets their own flags, or does not set flags and does not expect any special flags in the final linking. 

The acute problem here is that ESMX on GNU was linking with the `underflow` detection, which for the model in question is not a problem, but now, in DEBUG mode, the execution bails. It was finally determined that this came from ESMX, and was not expected.

theurich:
@danrosen25 I understand your motivation. However, the problem with making proactive decisions for the user (e.g. setting specific compiler run-time checks, etc. for their application) is that what works for some will not work for others. And if it is hidden inside the system (and cannot be affected easily by the user), it becomes a showstopper for some (e.g. NRL - which are still skeptical of ESMX, and magically bringing in compiler flags that affects their executable is a HUGE deal for them, so much that they'd probably ditch ESMX and just build their own application layer again - which I really want to avoid!!!). 

Specifically for the underflow here, yes, I think it is correctly caught by the GNU compiler, i.e. there is indeed an underflow in the calculation, i.e. something small is divided by something orders of magnitude larger. The underflow essentially results in a zero, but that is okay for the continuing calculation. However, with underflow detection active, this bails now. Thus DEBUG with underflow detection active is useless for them in this case, so they want it removed. However, since this was triggered, they are now aware that when building with ESMX there are these "magic" flags being set, and understandably have issues with that approach.

I guess we could have an `application` level switch in `esmxBuild.yaml` that lets a user turn on/off ESMX automatic flags for DEBUG. However, we only have those defined for GNU, Intel, and NVHPC anyway. Plus we don't guarantee how or when we update them, or what they exactly are. After some discussion and thought today, I am now also convinced that it's better for ESMX to NOT set "magic" debug flags for the final executable.

danrosen25:
That's what I'd like to see. You can remove the CMAKE_COMPILER_FLAGS and replace it with an option for ESMX debug modes. I would add an application level setting in ESMX that sets various levels of debugging as done now. This PR can be approved once a new ticket is opened.

theurich:
@danrosen25 please see #402.

danrosen25 (approved):

Issue #277, opened 2024-07-31
ESMF_RUNTIME_ environment variable to affect garbage collection behavior


theurich:
Currently some of the garbage collection code that runs during ESMF_Finalize() or ESMC_Finalize() is commented out. An idea that came out of the discussion #269 is to introduce a run-time environment variable `ESMF_RUNTIME_GARBAGE_EMPTY`, which if set would activate the currently commented out parts of the garbage collection code. This would allow running codes with full garbage collection enable to:

- Ensure no memory leaks are present. Something that some codes want to test for.
- Easily test situations for which the full garbage collection code is causing issues.

### Discussed in https://github.com/orgs/esmf-org/discussions/269

<div type='discussions-op-text'>

<sup>Originally posted by **AbhiHampiholi** July 25, 2024</sup>
### Requirements

- [X] Reviewed [ESMF Reference Manual](https://earthsystemmodeling.org/doc/)
- [X] Searched [GitHub Discussions](https://github.com/orgs/esmf-org/discussions?discussions_q=)

### Affiliation(s)

AirMettle, Inc

### ESMF Version

8.6.0

### Issue

Hello, 
In my code, all the allocated ESMC objects are reported as memory leaks despite calling the ESMC destroy functions and the Finalize function at the end. 
I noticed that the call to `ESMC_GridDestroy` calls the `ESMCI::Grid::destroy(&gridp)` which has signature `
  static int destroy(Grid **grid, bool noGarbage=false);` since no second argument is passed, the noGarbage flag is always false by default and inside the destroy function the `delete` is always skipped
  ```
    // optionally delete the complete object and remove from garbage collection
  if (noGarbage){
    VM::rmObject(*gridArg); // remove object from garbage collection
    delete (*gridArg);      // completely delete the object, free heap
  }
  ```
  I've attached a reproducible example here (copied from the example @danrosen25 sent me yesterday) [test_memleak.tar.gz](https://github.com/user-attachments/files/16380588/test_memleak.tar.gz)

  `valgrind --leak-check=full ./exeMemLeak` reveals the memory leaks. Let me know if I am missing something or if this is not-reproducible. If there isn't anything wrong with my setup, the fix might be to provide a way to set `noGarbage=true` while calling `ESMCI::Grid::destroy()` but I am not sure.  
 
 </div>

anntsay:
Gerhard - currently waiting on the PR #361  related to runtime variable to go through before finalize this.

anntsay:
[...]

PR #409, opened 2025-05-14
Control of ESMF garbage collection via ESMF_RUNTIME_ environment vars

theurich:
This PR addresses #277 by introducing two new ESMF_RUNTIME_ environment variables that provide user level control over the ESMF garbage collection functionality:

- `ESMF_RUNTIME_GARBAGE` with supported options `NONE`, `FULL`, and `SAFE` (default).
- `ESMF_RUNTIME_GARBAGE_LOG` with supported options `ON` and `OFF` (default).

Both variables are integrated into the standard `ESMF_RUNTIME_` environment handling, i.e. can also be set via configuration file and ingested during `ESMF_Initialize()`.

The default option, `ESMF_RUNTIME_GARBAGE=SAFE`, is implemented to provide unchanged garbage collection behavior compared to before this PR. This is the recommended way for typical scenarios.

The `ESMF_RUNTIME_GARBAGE=NONE` option disables the ESMF garbage collection. In this mode, the destroy responsibility for ESMF objects lies completely with the user. This can be useful when testing for memory leaks, e.g. for simple Create()/Destroy() code patterns. Notice however that this mode cannot provide double destroy and dangling alias protection!

The `ESMF_RUNTIME_GARBAGE=FULL` option enables all of the ESMF garbage collection code, including those portions that are not operating under `SAFE` mode. This might lead to potential crashes or undefined behavior in complex applications with coupled components accessing shared objects (potentially proxy objects). 

When experimenting with the ESMF garbage collection functionality, it can be helpful to set `ESMF_RUNTIME_GARBAGE_LOG=ON`. This instructs the garbage collection code to log information to the ESMF default log.

### DIFF
--- a/build_config/Linux.gfortran.default/build_rules.mk
+++ b/build_config/Linux.gfortran.default/build_rules.mk
@@ -132,13 +132,24 @@ ESMF_CCOMPILER_VERSION      = ${ESMF_CCOMPILER} --version
 ############################################################
 # Special debug flags
 #
-# Activate to turn on UBSan:
-#ESMF_LINKOPTFLAG_G      += -fsanitize=undefined
-# Also set environment variable UBSAN_OPTIONS="print_stacktrace=1"
-# for stacktrace at runtime.
-#
 ESMF_F90OPTFLAG_G       += -Wall -Wextra -Wconversion -Wno-unused -Wno-unused-dummy-argument -fbacktrace -fimplicit-none -fcheck=all,no-pointer
-ESMF_CXXOPTFLAG_G       += -Wall -Wextra -Wno-unused $(ESMF_LINKOPTFLAG_G)
+ESMF_CXXOPTFLAG_G       += -Wall -Wextra -Wno-unused
+
+############################################################
+# Special sanitizer flags
+#
+# Activate to turn on UBSan:
+#ESMF_OPTFLAG_G          += -fsanitize=undefined
+#ESMF_LINKOPTFLAG_G      += -fsanitize=undefined
+# Also set environment variable UBSAN_OPTIONS="help=1" to see available
+# run-time options.
+#
+# Activate to turn on ASan:
+#ESMF_OPTFLAG_G          += -fsanitize=address
+#ESMF_LINKOPTFLAG_G      += -fsanitize=address
+# Also set environment variable ASAN_OPTIONS="help=1" to see available
+# run-time options.
+#
 
 ############################################################
 # Fortran symbol convention
--- a/src/Infrastructure/VM/src/ESMCI_VM.C
+++ b/src/Infrastructure/VM/src/ESMCI_VM.C
@@ -1158,19 +1158,20 @@ void VM::shutdown(
               if (ESMC_LogDefault.MsgFoundError(localrc, ESMCI_ERR_PASSTHRU,
                 ESMC_CONTEXT, rc))
                 return;
-#if 0
-//gjt: Disable cleaning up LocStream from the garbage collection level because
-//gjt: it leads to issues when a field tries to destroy it later.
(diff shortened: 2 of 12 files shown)


build_config/Linux.gfortran.default/build_rules.mk
```diff
@@ -132,13 +132,24 @@ ESMF_CCOMPILER_VERSION      = ${ESMF_CCOMPILER} --version
...
+# run-time options.
+#
+# Activate to turn on ASan:
+#ESMF_OPTFLAG_G          += -fsanitize=address
+#ESMF_LINKOPTFLAG_G      += -fsanitize=address
+# Also set environment variable ASAN_OPTIONS="help=1" to see available
+# run-time options.
 #
-ESMF_F90OPTFLAG_G       += -Wall -Wextra -Wconversion -Wno-unused -Wno-unused-dummy-argument -fbacktrace -fimplicit-none -fcheck=all,no-pointer
-ESMF_CXXOPTFLAG_G       += -Wall -Wextra -Wno-unused $(ESMF_LINKOPTFLAG_G)
```
danrosen25, line 141: I assume removing `ESMF_LINKOPTFLAG_G` is collateral cleanup?

src/Infrastructure/VM/src/ESMCI_VM.C
```diff
@@ -11,7 +11,6 @@
 //==============================================================================
 #define ESMC_FILENAME "ESMCI_VM.C"
 //==============================================================================
-#define GARBAGE_COLLECTION_LOG_off
```
danrosen25, line 14: YES! I'm so happy to see preprocessor macros removed if they are not architecture/compiler/library dependendent.

src/Infrastructure/VM/src/ESMCI_VM.C
```diff
@@ -1179,50 +1202,83 @@ void VM::shutdown(
...
+            if (GlobalVM->garbageMode == garbageFull) mustDelete = true;
+            if (garbageLog){
+              char msg[800];
+              const char *proxyString;
+              proxyString="actual";
+              if (matchTable_Objects[i][k]->ESMC_BaseGetProxyFlag()==ESMF_PROXYYES)
+                proxyString="proxy";
+              const char *action;
+              action="KEEP";
+              if (mustDelete) action="DELETE";
```
danrosen25, line 1247: This is fine but slightly more efficient as an if...else

danrosen25 (approved):
Looks good. I see that you alphabetically ordered the ESMF_RUNTIME_ options so let's pull this first and then my PR, which I'll do the same. I'm missing documentation in the ESMX/README.md file for mine anyway. As someone not familiar with all the intricacies of garbage collection I only reviewed ESMCI_VM.C for any visual inconsistencies.

build_config/Linux.gfortran.default/build_rules.mk
theurich: Yes, I made handling of common link flags a bit more general, and no need for this implicit way of sneaking in `ESMF_LINKOPTFLAG_G` into `ESMF_CXXOPTFLAG_G`.

src/Infrastructure/VM/src/ESMCI_VM.C
theurich: Same here.

src/Infrastructure/VM/src/ESMCI_VM.C
theurich: Agree. Also seeing this again myself, I would have at least contracted lines 1245+1246 into `const char *action="KEEP";`. This came after moving some things around, so this is what it ended up no. Let's just keep it like this for now.

Issue #367, opened 2025-03-24
ESMF and ESMPy 8.8.0 versions differ (8.8.0 vs. 8.8.0b0), raising a VersionWarning


yantosca:
Not sure if this is a bug but I thought I'd report it.  Having installed the latest ESMF 8.8.0, I get a VersionWarning that the ESMF and ESMPy versions are different.  It seems like the ESMPy version is beta, is this still true?  Or was the version number not updated?  I didn't see anything in the issue tracker related to this.

```console
$ python -m gcpy.examples.plotting.create_test_plot
/home/bob/mambaforge/envs/dask_13_env/lib/python3.13/site-packages/esmpy/interface/loadESMF.py:94: VersionWarning: ESMF installation version 8.8.0, ESMPy version 8.8.0b0
  warnings.warn("ESMF installation version {}, ESMPy version {}".format(
```

I know I can squelch the warnings manually but it is a bit annoying to have to do so.

To replicate the issue you can download the geoschem-gcpy package from conda-forge

```console
$ mamba install geoschem-gcpy
$ export MPLBACKEND=tkagg 
$ python -m gcpy.examples.plotting.create_test_plot
```

Thanks so much!

DWesl:
The only thing I see is:
https://github.com/esmf-org/esmf/blob/bf628cb87505c2fc6892804b1a3054fd69e3d3c7/src/addon/esmpy/pyproject.toml#L65

Did you install from a tarball or a git checkout?  If tarball, where did you get it?

DWesl:
>   - esmf           ==8.8.0      # Earth System Model Framework
>   - esmpy          ==8.8.0      # Python wrapper for ESMF

This downloads ESMPy 8.8.0 as well.

The build setup is in [a conda-forge feedstock here](https://github.com/conda-forge/esmpy-feedstock/blob/main/recipe/meta.yaml), which downloads the GitHub-generated release tarball, [drops the versioning plugin from the runtime dependencies](https://github.com/conda-forge/esmpy-feedstock/blob/main/recipe/0001-remove-setuptools-git-versioning.patch), and uses Pip to build and install ESMPy.

Given the configuration for the versioning plugin at ESMF v8.8.0:
https://github.com/esmf-org/esmf/blob/cc31a2373cda3896f4ef902332224a88a60427ee/src/addon/esmpy/pyproject.toml#L35-L40
[...]

PR #415, opened 2025-05-18
DEV: Change python git version plugin to setuptools_scm.

DWesl:
Hopefully this will cause fewer problems than setuptools-git-versioning on builds from tarballs (e.g. esmf-org/esmf#367).

Behavior on git tags should be identical. I think I changed the behavior on commits after tags and on working trees with changes.  

It should be able to get the version from GitHub's auto-generated release/tag archives (https://github.com/esmf-org/esmf/archive/refs/tags/v8.8.1.tar.gz), it might be able to get a version from GitHub's auto-generated branch archives (https://github.com/esmf-org/esmf/archive/refs/heads/develop.zip), it keeps the existing fallback version in case neither of those worked, and downstream users should be able to set the `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_esmpy` environment variable if all of the previous attempts got borked somehow.

Descriptions of fallbacks [here](https://setuptools-scm.readthedocs.io/en/stable/usage/#builtin-mechanisms-for-obtaining-version-numbers) and [here](https://setuptools-scm.readthedocs.io/en/stable/config/).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,1 @@
-src/doc		export-ignore
-.gitattributes	export-ignore
-.gitignore	export-ignore
-
-*.C linguist-language=C++
+.git_archival.txt         export-subst
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,7 @@
-.git_archival.txt         export-subst
+src/doc		export-ignore
+.gitattributes	export-ignore
+.gitignore	export-ignore
+
+*.C linguist-language=C++
+
+.git_archival.txt export-subst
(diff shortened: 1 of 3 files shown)


billsacks:
As per [my comment in this discussion](https://github.com/orgs/esmf-org/discussions/352#discussioncomment-13349381), I'd like to suggest the changes in #423 in place of this. @DWesl I really appreciate your looking into this and suggesting a possible path forward with setuptools-scm... that does seem like a better approach than the existing one if the changes in #423 end up having other unforeseen problems. @DWesl, since you've given this a fair amount of thought and experimentation, I'd welcome any thoughts you have on this.

billsacks:
Based on [conversation in #352](https://github.com/orgs/esmf-org/discussions/352#discussioncomment-13372591), I'm going to close this and we'll go with #423 for now. It's great to have the changes here as a fall-back option, though, if we run into any problems with #423. Thanks again for your contribution and thoughts, @DWesl !

PR #422, opened 2025-05-30
Fix ESMF CI for macos-14 where libgfortran not loaded

danrosen25:
Tested with linking libgfortran into usr/local/lib after creating usr/local/lib and it was occasionally successful
Copied setup from gfortran to gfortranclean and forced cache rebuild and it was successful
Reasoning for the second test is that the gfortran build continued to work after the macos upgrades.


danrosen25:
Apparently this wasn't the fix because it ran here: https://github.com/danrosen25/esmf/actions/runs/15355374822
And now I'm not sure why it worked versus didn't work.

danrosen25:
Actually, I do have an idea. Let me try deleting the cache.

Issue #420, opened 2025-05-30
Incorrect ESMF_VERSION_STRING_GIT for ESMF forks missing tags


danrosen25:
ESMF_VERSION_STRING_GIT is based on `git describe --tag` and this command is incorrect in forked repositories that have not copied tags from esmf-org. Tags can be copied to a fork using git fetch --tags https://github.com/esmf-org/esmf.git. This also affects the version calculation in esmpy, which is the bigger problem.

danrosen25:
Hi @billsacks @theurich 
I just wanted to bring up this issue with you because at some point we need to fix this in esmpy at the least. My [ESMF fork](https://github.com/danrosen25/esmf) only has tags for 8.6.0 and 8.6.1 so the esmpy build was picking that up in the ESMF_VERSION_STRING_GIT. This was causing an inconsistency and failure. Do you think we should avoid using ESMF_VERSION_STRING_GIT? Do we need a better way of versioning the entire ESMF repository in one place?

theurich:
I am not sure that this is in the release procedure wiki. My convention was to create the first tag of a release cycle, i.e. `vX.Y.0b00`, as annotated tag. E.g. see https://github.com/esmf-org/esmf/releases/tag/v8.8.0b00. This way `git describe` should at least reference no further back than the current start of the release cycle.

I think it's actually okay to retag `v8.9.0b00`, since it would keep the tag at the exact same place, just be annotated instead of light.

oehmke:
Since we’ve been inconsistent about it, should we just leave things as they are (e.g. with 8.9.0b00)? We can add something to the release procedure and do it going forward. Or do you want to change 8.9.0b00?

theurich:
Let's see what @danrosen25 gets for `git describe` under his fork right now... then if we come up with a fall-back solution for `ESMF_VERSION_STRING_GIT` that involves the annotated tags, we might want to "fix" `v8.9.0b00`. If not, then we can just leave it, and as you say, do it going forward.

danrosen25:
@theurich 
I'd also be okay with just concatenating ESMF_VERSION_STRING with the shortened git hash and calling it a day. That does exactly what it says and will give all the information we need because the hash will be included. I don't think we need the comparison logic.

billsacks:
@danrosen25 - thanks for starting this discussion. I'm still trying to understand the scenarios where this is a problem... so sharing a few tentative thoughts here:

[...]

PR #423, opened 2025-06-02
Get ESMPy version from the Macros file instead of the git version

billsacks:
We have run into various issues with trying to get the version from the git version - e.g.:
- https://github.com/esmf-org/esmf/issues/420
- https://github.com/orgs/esmf-org/discussions/352
- https://github.com/esmf-org/esmf/issues/367
- https://github.com/esmf-org/esmf/pull/415

Per discussion in https://github.com/esmf-org/esmf/issues/420, we feel it's best to stop trying to get the version from the git tag and instead use the same Macros file that's used to get the version for other purposes in ESMF. That's what's done here.

The main behavior change I can think of is that, for development tags, we will no longer have the exact beta tag. Instead, this will always be given as something like "8.9.0b0" unless we update ESMC_Macros.h for each beta tag. (It seems that *some* beta tag number is required, so I don't think we can get simply "8.9.0beta" or "8.9.0b": setuptools-git-versioning automatically translates something like "8.9.0 beta" to "8.9.0b0".)


billsacks:
Thanks, @danrosen25 - I had a typo in a path that worked on my Mac but not on Linux since Linux is case-sensitive and MacOS isn't... I've fixed that and hopefully that will resolve the problem.

src/addon/esmpy/build_utils/version.py
danrosen25, line 1: Needs "newline" terminator at end of file.

PR #426, opened 2025-06-04
Append git origin to ESMF_VERSION_STRING_GIT

danrosen25:
Appends the git remote origin to ESMF_VERSION_STRING_GIT
If no origin is found then reports no origin found
Also fixes an issue where if no tag is found then ESMF assumes it is not a git repository. This is common when using GitHub actions/checkout because the fetch depth is zero.

Previous make info results
```
--------------------------------------------------------------
ESMF_VERSION_STRING:      8.9.0 beta snapshot
Not a Git repository
--------------------------------------------------------------
```

New make info results
```
--------------------------------------------------------------
ESMF_VERSION_STRING:      8.9.0 beta snapshot
ESMF_VERSION_STRING_GIT: no tag found (https://github.com/danrosen25/esmf)
--------------------------------------------------------------
On branch version_string_git
Your branch is up to date with 'origin/version_string_git'.

nothing to commit, working tree clean
--------------------------------------------------------------
```

### DIFF
--- a/scripts/esmfversiongit
+++ b/scripts/esmfversiongit
@@ -2,5 +2,12 @@
 # return the ESMF version from Git if available, or empty string otherwise
 # test looks for .git directory or .git file (when esmf is a submodule)
 if [ -d $ESMF_DIR/.git ] || [ -s $ESMF_DIR/.git ] ; then \
-git describe --tags 2>&1 | grep -v fatal
+TAG=`git describe --tags 2>&1 | grep -v fatal`
+ORIG=`git config --get remote.origin.url 2>&1`
+if [ -z "$TAG" ] ; then TAG="no tag found" ; fi
+if [ -z "$ORIG" ] ; then ORIG="no origin found" ; fi
+case "$ORIG" in
+  *"esmf-org/esmf.git"*) echo $TAG ;;
+  *) echo $TAG "($ORIG)" ;;
+esac
 fi


theurich:
I think the following block of code will need to be updated to work with the approach: https://github.com/danrosen25/esmf/blob/22d577205f8dcf4819cd8c95d97f987330502fcc/src/Superstructure/ESMFMod/src/ESMF_Init.F90#L1211-L1227

danrosen25:
> I think the following block of code will need to be updated to work with the new approach: https://github.com/danrosen25/esmf/blob/22d577205f8dcf4819cd8c95d97f987330502fcc/src/Superstructure/ESMFMod/src/ESMF_Init.F90#L1211-L1227

@theurich Oh I see. ESMF_VERSION_STRING_GIT won't change in the case of origin=esmf-org/esmf.git AND depth>tag depth. But, there are a lot of workflows out there where depth=1 and they will now report no tag found, which means this block of code now uses ESMF_VERSION_STRING_GIT instead of ESMF_VERSION_STRING.

I have to think on it.

danrosen25:
@theurich @billsacks
I've tried to think about this a bit and I'm not sure what information we want. I kind of like the following. This would report a tag if the specific commit is tagged. If the commit is not tagged then it will print the branch+sha1. It will also append the origin information when it's not esmf-org/esmf. it gives you all the git information needed to locate the source but simplifies it to a single tag if it can.

```sh
if [ -d $ESMF_DIR/.git ] || [ -s $ESMF_DIR/.git ] ; then \
TAG=`git describe --tags --exact 2>&1 | grep -v fatal`
BRCH=`git rev-parse --abbrev-ref HEAD 2>&1`
SHA1=`git rev-parse --short HEAD 2>&1`
if [ -z "$TAG" ] ; then TAG="$BRCH+g$SHA1" ; fi
ORIG=`git config --get remote.origin.url 2>&1`
case "$ORIG" in
  *"esmf-org/esmf"*) echo $TAG ;;
  *) echo $TAG"#"$ORIG ;;
esac
fi
```

billsacks:
I like that, @danrosen25 . My understanding is that, if you want to know the github source for the current commit, this version assumes that it exists on origin (as opposed to on some other remote). You could determine this more definitively with something like:

```
git branch -r --contains HEAD | cut -d'/' -f1 | sort -u
```

which gives you a list of all remotes containing the current commit. But then you'd need to get the url for each remote, turn it into something like a comma-delimited list, and... that's starting to feel pretty complex for handling this edge case... so I feel like what you have now is great until we find a need for something more.

PR #429, opened 2025-06-05
Minor fix for spack build

uturuncoglu:
It seems that there are some changes in spack side and `spack compiler list` command output format slightly changed and now includes extra information like `[e]` at the beginning. 

```
root@af01e8d67ed4:/opt# spack compiler list
==> Available compilers
-- gcc ubuntu24.04-aarch64 --------------------------------------
[e]  gcc@13.3.0
```

So, this PR modifies script to accommodate to the recent format change.


theurich (approved):
Always something... Thanks for fixing this so quickly!

PR #435, opened 2025-06-15
Enable multiple DEs per PET in ESMPy

billsacks:
Extend the C and Python APIs to support multiple DEs per PET in ESMPy.

The main current use case is working with a cubed-sphere grid with fewer than 6 processors.


billsacks:
This is a work-in-progress, but I am opening a draft PR so that others can more easily give feedback on pieces of it.

billsacks:
@oehmke or @theurich - I'd like a quick sanity check from one of you on [this commit](https://github.com/esmf-org/esmf/pull/435/commits/95c5d6f2574d98d843658ea16a9b77a7e2313564). I wanted to add a C API function to get the local DE count for a Field (in addition to the function I added earlier to get the local DE count for a Grid). It seemed like the way to do this is to get the local DE count for the Array associated with the Field. So I figured that I might as well add a C API function to get a local DE count for an Array while I was at it. Does this feel reasonable - or, for example, do we want to keep the C API for Array more bare-bones (in which case I can rework this to avoid going through a C API function for the Array, instead going into the C++)?

Also note that, since the Field API invokes the Array API, I didn't bother adding a unit test for the Array API separate from the one for the Field API.

oehmke:
It seems fine to me to have a method to get the local DE count from an Array. It seems like the natural way to do things and I can’t think of a big downside to having the extra functionality in Array.

billsacks:
Finally, the culminating change of this PR: e63f8bbe3546113634e637580768112b5c5f18b7. This commit is small but mighty, leveraging the changes I made up to this point in the PR. There are three important parts to this commit:
1. Add a loop over DEs, working on one DE at a time in getting coordinates and setting field data
2. Uncomment assertions on the regridding error. Note that, before the changes in this PR, these assertions were failing even with 6 PETs because the decomposition in this example is set up so that there are multiple DEs per PET even with 6 PETs.
3. Run this test with any number of PETs; it now works (including the final assertions) with any number of PETs

billsacks:
In looking for other things that may need to be changed, I noticed that this function would need some work to work properly with multiple DEs per PET:

https://github.com/esmf-org/esmf/blob/6f30baa24709cbd3c55e379667e27e7433ffc428/src/addon/esmpy/src/esmpy/util/field_utilities.py#L11

Currently it should fail with an appropriate exception if someone tries to use it with multiple DEs per PET. However, as far as I can tell, this function isn't a documented part of the ESMPy API, and from a quick look it appears to only be used internally in unit tests (either directly or via regrid_check). So fixing this doesn't feel high priority to me - and, at least, shouldn't hold up the upcoming release.

PR #437, opened 2025-06-16
Fix compiling on Cray with CCE 19.0.0

jvgeiger:
Compiling on Cray with the 19.0.0 compilers raises this error:

```
ESMCI_MeshCapRegridUTest.C:94:26: error: implicit instantiation of undefined template 'std::array<char, 64>'
   94 |     std::array<char, 64> buffer;
      |                          ^
```

This is resolved by adding "#include <array>".


theurich (approved):
Looks good.

PR #440, opened 2025-06-17
Support the traditional CMake build procedure with ESMX (broken for some cases)

theurich:
There is an edge case for which the traditional CMake build procedure was broken under ESMX. For the traditional CMake procedure, the configuration step is executed from within the `build` directory as `cmake ..`, or pointing to wherever the `CMakeLists.txt` file is located. The traditional approach is expected to work with ESMX in addition to the modern CMake approach, using the `-S` and `-B` options, and also the `ESMX_Builder` approach.

The problem with the traditional CMake approach comes in when ESMX is searching for component source directories, where the default source directory is assumed to be a subdirectory under the current working directory. However, for the traditional CMake procedure that will be inside of `build`, where of course component sources are NOT located. The fix implemented in this PR is to use the `CMAKE_SOURCE_DIR`, which is the expected default. This works for both traditional and modern CMake procedures. To be compatible with `ESMX_Builder`, pass the current working directory (from where ESMX_Builder is expected to be called) into CMake via variable.

The solution was tested with the NUOPC apps protos, and also under NEPTUNE and ESPC, where the edge case was found. The PR resolves the issue.

### DIFF
--- a/src/addon/ESMX/Driver/CMakeLists.txt
+++ b/src/addon/ESMX/Driver/CMakeLists.txt
@@ -65,6 +65,11 @@ else()
   set(ESMX_EXE_NAME ${ESMX_EXE_NAME} PARENT_SCOPE)
 endif()
 
+# define ESMX_SDIR
+if(NOT DEFINED ESMX_SDIR)
+  set(ESMX_SDIR ${CMAKE_SOURCE_DIR})
+endif()
+
 # test options
 if(ESMX_TEST)
   enable_testing()
@@ -133,7 +138,7 @@ endif()
 
 # generate component config file(s)
 execute_process(COMMAND ${Python_EXECUTABLE}
-  ${CMAKE_CURRENT_LIST_DIR}/esmx_comp_config.py --ifile ${ESMX_BUILD_FILE} --odir ${CMAKE_CURRENT_BINARY_DIR} --sdir ${CMAKE_SOURCE_DIR} ${ESMX_DISABLE_COMPS}
+  ${CMAKE_CURRENT_LIST_DIR}/esmx_comp_config.py --ifile ${ESMX_BUILD_FILE} --odir ${CMAKE_CURRENT_BINARY_DIR} --sdir ${ESMX_SDIR} ${ESMX_DISABLE_COMPS}
   RESULT_VARIABLE ret)
 esmx_check_ret(${ret} "esmx_comp_config.py failed processing ${ESMX_BUILD_FILE}")
 
--- a/src/addon/ESMX/Driver/CMakeLists.txt
+++ b/src/addon/ESMX/Driver/CMakeLists.txt
@@ -133,7 +133,7 @@ endif()
 
 # generate component config file(s)
 execute_process(COMMAND ${Python_EXECUTABLE}
-  ${CMAKE_CURRENT_LIST_DIR}/esmx_comp_config.py --ifile ${ESMX_BUILD_FILE} --odir ${CMAKE_CURRENT_BINARY_DIR} ${ESMX_DISABLE_COMPS}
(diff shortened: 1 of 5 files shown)


theurich:
@danrosen25 and @oehmke unfortunately I was not able to be there for the whole core call this morning... and probably you all touched on the 8.9.0 freeze. I am really hoping to get this PR still into 8.9.0... It passes checks, and also just ran the Spack test action on this branch https://github.com/esmf-org/esmf/actions/runs/15744948186, which tests the NUOPC app protos. All looking good. Let me know if you think this needs more time for review. Otherwise I'd like to get this in today. Thanks.

src/addon/ESMX/Driver/esmx_comp_config.py
```diff
@@ -43,7 +43,7 @@ def create_compList(cmpCfg: ESMXCmpCfg, odir):
                         for i in range(len(dirs)):
                             dirs[i] = dirs[i].strip()
                             if not dirs[i].startswith('$'):
-                                dirs[i] = os.path.abspath(dirs[i])
+                                dirs[i] = os.path.abspath(sdir + "/" + dirs[i])
```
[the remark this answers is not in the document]
theurich: I think you are correct. Good catch! Do you have a concrete code suggestion, we could review within this comment thread.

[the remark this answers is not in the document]
danrosen25: Maybe this?
```python
if dirs[i].startswith('/'):
  dirs[i] = os.path.abspath(dirs[i])
elif not dirs[i].startswith('$'):
  dirs[i] = os.path.abspath(sdir + "/" + dirs[i])
```

src/addon/ESMX/Driver/CMakeLists.txt
```diff
@@ -65,6 +65,11 @@ else()
   set(ESMX_EXE_NAME ${ESMX_EXE_NAME} PARENT_SCOPE)
 endif()
 
+# define ESMX_SDIR
+if(NOT DEFINED ESMX_SDIR)
+  set(ESMX_SDIR ${CMAKE_SOURCE_DIR})
```
danrosen25, line 70: I'm trying to figure out if this makes sense or if it needs to be `set(ESMX_SDIR ${CMAKE_CURRENT_SOURCE_DIR})` or `set(ESMX_SDIR ${CMAKE_CURRENT_LIST_DIR})` or `get_filename_component(ESMX_SDIR ${ESMX_BUILD_FILE} DIRECTORY)`. The last one uses the directory where the esmxBuild file is located.

danrosen25: `${CMAKE_CURRENT_SOURCE_DIR})` might also satisfy your need while allowing the ESMX build to be embedded deeper into a cmake build, say for a build that has multiple applications. Do you think the ESMX build will always look for the source from the root CMAKE source directory or from the directory calling the ESMX build. ESMX_Builder is more similar to the latter, which assume the sources are in the directory in which the ESMX build is called.

theurich: The `CMAKE_CURRENT_SOURCE_DIR` will point to the directory that contains the ESMX Driver `CMakeLists.txt` because it is hooked into the application CMake lists file with `add_subdirectory()`. The case you are describing of using ESMX to build an application within an even larger CMake based build system is still achievable by setting the `ESMX_SDIR` variable explicitly to `CMAKE_CURRENT_LIST_DIR` from the level that contains the `add_subdirectory()` for ESMX.
Bottom line, I think what's in the PR now is the correct way. But it's nice to have thought through all of these scenarios!

danrosen25 (approved):

PR #442, opened 2025-06-24
Fix a problem that trips up new version of CCE compiler

theurich:
This PR fixes an issue in the ESMF Trace implementation which under single PET execution would lead to undefined behavior. However, on Cray's CCE version 19, in optimized mode incorrect code is generated, leading to hanging unit tests. 

This PR fixes https://github.com/esmf-org/esmf-support/issues/543.

### DIFF
--- a/src/Infrastructure/Trace/src/ESMCI_TraceClock.C
+++ b/src/Infrastructure/Trace/src/ESMCI_TraceClock.C
@@ -186,8 +186,8 @@ namespace ESMCI {
 
     int myPet = vm->getLocalPet();
     int petCount = vm->getPetCount();
-    int64_t ret;
-    
+    int64_t ret = 0;  // initialize to ensure valid return for all conditions
+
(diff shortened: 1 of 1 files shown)


danrosen25:
@theurich 
Did you also see this PR https://github.com/esmf-org/esmf/pull/437?

theurich:
> @theurich Did you also see this PR #437?

@danrosen25 thanks for pointing this out. I had not noticed that PR. I have since also run into the issue with the include missing (when I ran more testing). I merged the PR from @jvgeiger into this PR's branch. Should be all good now.

Sorry @jvgeiger if you started building running already with this PR's branch before your changes were merged in. If so, please pull again to make sure you have the most up-to-date version before testing. Thank you!

theurich:
Pending results from @jvgeiger running again with `ESMF_OPTLEVEL=1` as to avoid CCE 19 issues when building the ESMF system tests. Once confirmed recommend merging into `develop` for 8.9.0 release.

PR #444, opened 2025-06-25
ESMPy: Fix Grid size property following slicing

### LINKED ISSUE #441: ESMPy Grids return incorrect `size` after slicing
When slicing an ESMPy Grid (e.g., `grid2 = grid[1:21, 3:17]`), the upper bounds of the new grid are set correctly, but the size is not. So an attempt to get the size (via `grid2.size` or `grid2.all_sizes`) will give the size of the original grid, not the sliced grid.

This error can be seen with the following additions to the grid unit test; the second assertion fails:

```diff
diff --git a/src/addon/esmpy/src/esmpy/test/test_api/test_grid.py b/src/addon/esmpy/src/esmpy/test/test_api/test_grid.py
index 213a541778..c162702315 100644
--- a/src/addon/esmpy/src/esmpy/test/test_api/test_grid.py
+++ b/src/addon/esmpy/src/esmpy/test/test_api/test_grid.py
@@ -306,11 +306,13 @@ def test_grid_slice_2d(self):
 
         assert grid.coords[StaggerLoc.CENTER][0].shape == (100, 100)
         assert grid.upper_bounds[StaggerLoc.CENTER].tolist() == [100, 100]
+        assert grid.size[StaggerLoc.CENTER].tolist() == [100, 100]
 
         del grid
 
         assert grid2.coords[StaggerLoc.CENTER][0].shape == (20, 14)
         assert grid2.upper_bounds[StaggerLoc.CENTER].tolist() == [20, 14]
+        assert grid2.size[StaggerLoc.CENTER].tolist() == [20, 14]
 
         del grid2
 ```

I feel like the best fix would be to stop storing the grid size at all (i.e., remove the `._all_sizes` variable), instead computing this on the fly when needed (via the `all_sizes` property, as the difference between `upper_bounds` and `lower_bounds`).

(I noticed this issue while working on #435, but have confirmed that the issue existed both before and after the changes for #435 .)

### PR DESCRIPTION
**Pull request by billsacks**
This PR fixes the size property of an ESMPy Grid created by slicing another Grid. This also adds tests of this case (which failed before these changes but pass now).

Note that I have implemented this by calculating the size property on the fly rather than storing it. This seems less error-prone, because now size doesn't need to be kept in sync with the bounds.

Resolves #441

### DIFF
--- a/src/addon/esmpy/src/esmpy/api/grid.py
+++ b/src/addon/esmpy/src/esmpy/api/grid.py
@@ -428,9 +428,6 @@ class Grid(object):
         # all_upper_bounds[de][staggerLoc]
         self._all_upper_bounds = [[None for _ in range(2**self.rank)]
                                   for _ in range(self.local_de_count)]
-        # all_sizes[de][staggerloc]
-        self._all_sizes = [[None for _ in range(2**self.rank)]
-                           for _ in range(self.local_de_count)]
         # all_coords[de][staggerLoc][coord_dim]
         self._all_coords = [[[None for _ in range(self.rank)]
                              for _ in range(2**self.rank)]
@@ -591,7 +588,11 @@ class Grid(object):
             numpy arrays of ints of size given by
             ``upper_bounds - lower_bounds``.
         """
-        return self._all_sizes
+        return [[(None if self.all_upper_bounds[localde][stagger] is None
+                  or self.all_lower_bounds[localde][stagger] is None
+                  else self.all_upper_bounds[localde][stagger] - self.all_lower_bounds[localde][stagger])
+                  for stagger in range(2 ** self.rank)]
+                  for localde in range(self.local_de_count)]
 
     @property
     def all_upper_bounds(self):
@@ -828,7 +829,7 @@ class Grid(object):
 
         if self.local_de_count > 1:
             raise SingleLocalDEMethod
-        return self._all_sizes[0]
+        return self.all_sizes[0]
 
     @property
     def staggerloc(self):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/addon/esmpy/src/esmpy/api/grid.py [hunk 1]
```diff
@@ -591,7 +588,11 @@ def all_sizes(self):
             numpy arrays of ints of size given by
             ``upper_bounds - lower_bounds``.
         """
-        return self._all_sizes
+        return [[(None if self.all_upper_bounds[localde][stagger] is None
+                  or self.all_lower_bounds[localde][stagger] is None
+                  else self.all_upper_bounds[localde][stagger] - self.all_lower_bounds[localde][stagger])
```
oehmke, line 593: @billsacks just looking at this quickly, but shouldn't it it be upper-lower+1? (e.g. 5 to 1 should be size 5 not 4), although I don't do a lot of Python, so maybe I don't see what's going on here.

billsacks: Very good question. You had me scratching my head about this for a bit. To be honest, I just moved that logic from elsewhere without giving it much thought, but your question inspired me to give it a little more thought / investigation. I'm still not *positive*, but based on some code I'm seeing in ESMPy, it looks like this is correct. It seems like this is following python indexing conventions, where bounds are 0-indexed and inclusive of the lower-bound but exclusive of the upper-bound. So, for example, x[0:3] gives 3 elements, at indices 0, 1 and 2.

This also explains this other bit of code that I remember being puzzled by but now makes sense:

https://github.com/esmf-org/esmf/blob/542024b59e7ee43ae0528793a111fafe2a791bc8/src/addon/esmpy/src/esmpy/interface/cbindings.py#L944-L945

(When I first looked at this, I wondered: doesn't this change the size? But now I see how this is all consistent with a switch to Python-style bounds.)

oehmke:
Oh, ok, that makes sense. I guess that it’s one of those language idioms. We’ll need to be careful with that as all if the bounds returned from the Fortran API include the upper value, but I guess that code you point to shows how that’s handled.  

Given that, then your change looks ok to me.

PR #445, opened 2025-06-27
Fix ORIG symbol in esmfversiongit

danrosen25:
Using `#` was a bad idea because it cuts off the remainder of a line in the esmf.mk makefile
Can I swap this to `:` or do you think there's a better symbol?

esmf_mk_example

### DIFF
--- a/scripts/esmfversiongit
+++ b/scripts/esmfversiongit
@@ -9,6 +9,6 @@ if [ -z "$TAG" ] ; then TAG="$BRCH+g$SHA1" ; fi
 ORIG=`git config --get remote.origin.url 2>&1`
 case "$ORIG" in
   *"esmf-org/esmf"*) echo $TAG ;;
-  *) echo $TAG"#"$ORIG ;;
+  *) echo $TAG":"$ORIG ;;
 esac
 fi


billsacks:
I'm happy with either `:` or `@`

danrosen25:
My concern is that the whatever symbol should always be treated like a text character and `@` is often specialized. Otherwise I liked `@` better too.

The compile line that failed looked like this
<code>mpifort -c -g -fPIC  -pthread -frecursive -ffree-line-length-none -fno-backtrace -I/Users/drosen/Utils/esmf/clang16-gfortran14/O/develop/mod -I/Users/drosen/Utils/esmf/clang16-gfortran14/O/develop/include -I/Users/drosen/Utils/netcdf/4.9.2_clang16-gfotran14/include -I/Users/drosen/Utils/netcdf/4.9.2_clang16-gfotran14/include  -DESMF_NO_INTEGER_1_BYTE -DESMF_NO_INTEGER_2_BYTE -DESMF_VERSION_STRING_GIT='develop+g344364f0d1 tawas.F90
/bin/sh: -c: line 0: unexpected EOF while looking for matching `''
/bin/sh: -c: line 1: syntax error: unexpected end of file</code>

This is because `include $(ESMFMKFILE)` failed on `ESMF_F90COMPILECPPFLAGS=` when it cut the line at the `#` symbol.

PR #446, opened 2025-06-28
Improve unit testing of multiple DEs per PET for ESMPy Grids

billsacks:
The changes here give better test coverage of the multiple-DE-per-PET case for ESMPy Grids, such as for coordinates, areas and masks.

This also adds some additional testing even for the single-DE-per-PET case - specifically, ensuring that the linkage between the ESMPy and ESMF mask and area data are set up correctly.

### DIFF
--- a/src/addon/esmpy/examples/cubed_sphere_to_mesh_regrid.py
+++ b/src/addon/esmpy/examples/cubed_sphere_to_mesh_regrid.py
@@ -46,6 +46,8 @@ dstfracfield = esmpy.Field(dstgrid, name='dstfracfield', staggerloc=esmpy.Stagge
 
 deg2rad = 3.14/180.
 
+# Since the source grid and field (on the cubed sphere grid) have multiple DEs per PET,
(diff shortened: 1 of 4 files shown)


src/addon/esmpy/examples/cubed_sphere_to_mesh_regrid.py
```diff
@@ -46,6 +46,8 @@
 
 deg2rad = 3.14/180.
 
+# Since the source grid and field (on the cubed sphere grid) have multiple DEs per PET,
+# working with the data on that grid/field requires looping over the local DEs on each PET
 for de in range(srcfield.local_de_count):
```
oehmke, line 51: @billsacks just wanted to double check that this does the range 0 to local_de_count-1

oehmke (approved):
I had one comment verifying that values produced by range(local_de_count) is 0 to local_de_count-1 However, if that's true (which I suspect it is), then it all looks good. At some point we should broaden the compare_fields() utility to be multiple  des, but that can be after the release. Thanks for all of your careful work on this!

PR #452, opened 2025-07-02
Add ESMPy testing on 8 PEs

billsacks:
This covers the situation where we have cubed-sphere grids with some PETs having 0 DEs.

I have confirmed that tests pass with 8 PEs on my Mac.

I have also tested with 10 PEs on my Mac and tests also pass with that PE count, but I feel like it probably isn't worth the testing time to add testing with both 8 and 10 PEs. (In fact, even adding this testing with 8 PEs seems like marginal benefit: As far as I'm aware, the only extra coverage it adds is to cover the issue fixed by https://github.com/esmf-org/esmf/pull/447/commits/523a99779d3ce8dd98b0df0614ec7ec26f4b0a6c.)

I ran into problems trying to run testing with 12 PEs on my Mac (note that I need `-oversubscribe` in the mpirun command for running testing with 12 PEs... it's possible that some issues were due to this, though I haven't experienced issues with oversubscribing processors in the past): some tests in test_regrid.py fail with this processor count (I haven't tried digging into why), and then I got a hang.

### DIFF
--- a/src/addon/esmpy/src/esmpy/test/test_all.bash
+++ b/src/addon/esmpy/src/esmpy/test/test_all.bash
@@ -21,7 +21,7 @@ if [ "$MPIEXEC_END_OF_PATH" = "mpiuni/mpirun" ]
 then
     PE_COUNTS=(1)
 else
-    PE_COUNTS=(1 4 6)
+    PE_COUNTS=(1 4 6 8)
 fi
 
 for NP in ${PE_COUNTS[@]}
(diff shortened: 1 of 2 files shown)


billsacks:
With the commit I just pushed, ESMPy tests pass with 12 and 16 PETs on my Mac. (This commit just skips some tests that don't seem to be reasonable tests on > 10 PETs.) (I ran into a failure with 18 PETs in `test_read_write_weight_file`, but I don't feel like it's worth the effort to dig further on this right now: my guess is that I'll find that this is due to a similar issue, i.e., the test case being too small for this PET count.)

I don't feel like it's worth adding regular testing of ESMPy with more than 8 PETs for now, but did want to document here that this is working.

oehmke:
That sounds good. I think that the set of PET counts that you are using now seems reasonable because it tests a thorough range of local DE count cases.

PR #455, opened 2025-07-07
Update ESMPy documentation for allowing multiple DEs per PET

billsacks:
This PR adds some documentation on the use of multiple DEs per PET in ESMPy and removes a no-longer-true limitation.


oehmke (approved):
Looks good to me. Thanks!

PR #462, opened 2025-07-10
Correctly handle absent regDecompPTile in ESMC_GridCreateCubedSphere

billsacks:
This is needed for ESMPy testing to pass on derecho-intel.

I see (at least) two methods for handling optional arguments passed from C to Fortran:
1. An explicit "foopresent" integer passed as an argument and then checked, with "foo" *not* declared as optional in the Fortran
2. Declaring "foo" as optional in the Fortran, with the C passing a NULL pointer

(2) seems like the cleaner approach where possible, and it avoids a combinatoric explosion of conditionals. But I'm using (1) here because that's what was set up. **Do we (eventually) want to migrate to (2) in cases like this, or is this not possible in this situation?**

Note that there is currently no combinatoric explosion in this routine because two of the arguments that should be optional (decompFlagPTile and deLayoutList) are currently commented-out. (Eventually I imagine we may want to enable those arguments....)

### DIFF
--- a/src/Infrastructure/Grid/interface/ESMF_Grid_C.F90
+++ b/src/Infrastructure/Grid/interface/ESMF_Grid_C.F90
@@ -208,14 +208,28 @@
     enddo
 
 
-    grid = ESMF_GridCreateCubedSphere(tilesize, &
-                                      regDecompPTile=regDecompPTile, &
-                                      !decompFlagPTile=decompFlagPTile_local, &
-                                      !deLabelList=deLabelList, &
-                                      !delayout=delayout, &
-                                      staggerLocList=staggerLocList_local, &
-                                      name=name, &
-                                      rc=rc)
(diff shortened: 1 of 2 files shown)


theurich:
There is also the official Fortran 2008 way to avoid combinatoric explosion with a local pointer variables for each optional argument passed on to Fortran. Initialize to `=>null()`, and point to actual present dummies as per guard variable coming in from C side, similar to what you describe under your option #1. The advantage over #2 is that it avoids combinatoric explosion while being completely standard Fortran.

For now though I suggest moving forward with what you are proposing.

theurich (approved):
Good to go.

billsacks:
Some derecho testing is still underway (https://earthsystemmodeling.org/esmf-test-summary/fix_esmpy_cubed_sphere/index.html), but all of the combos that exercise ESMPy are finished and all ESMPy tests now pass, so I'm going ahead and merging this so it will be in place for tonight's nightly testing.

PR #463, opened 2025-07-10
Fix language in Attribute deprecation statement

theurich:
This PR replaces the specific 9.0.0 target with more vague language to be consistent with next release being 9.0.0, and Attribute work is not scheduled.

### DIFF
--- a/src/Superstructure/AttributeAPI/doc/Attribute_options.tex
+++ b/src/Superstructure/AttributeAPI/doc/Attribute_options.tex
@@ -2,8 +2,8 @@
 
 \subsubsection{ESMF\_ATTCOPY}
 \label{const:attcopy}
-\apiDeprecatedClassWithTargetReleaseAndReplacement{ESMF\_Attribute}{9.0.0}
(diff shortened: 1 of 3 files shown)


uturuncoglu (approved):
It looks good to me.

billsacks:
Copying an email to here:

What about changing "is scheduled for removal with" -> "will be removed in"? I think that works with either something specific ("will be removed in ESMF 9.0.0") or something more vague ("will be removed in a future release").

But I'm also fine with just sticking with what you have here.

theurich:
> Copying an email to here:
> 
> > What about changing "is scheduled for removal with" -> "will be removed in"? I think that works with either something specific ("will be removed in ESMF 9.0.0") or something more vague ("will be removed in a future release").
> 
> But I'm also fine with just sticking with what you have here.

I like your suggestion wrt wording ... will update push the change under this PR.

billsacks (approved):
Looks good! Thanks for thinking of this and making this change!

PR #467, opened 2025-07-24
Prevent floating point exceptions (FPE) due to compiler optimization pre-computation

### LINKED ISSUE #464: Bug involving ESMF 8.9, ifort 2021.13, and fpe0
We seem to have encountered an issue with ESMF 8.9 involving Intel ifort 2021.13 and building ESMF code with `-fpe0`. We first saw this in MAPL, but @bena-nasa was able to make a reproducer:

[ESMF89-Bug-Reproducer.tar.gz](https://github.com/user-attachments/files/21239978/ESMF89-Bug-Reproducer.tar.gz)

This reproducer is (at the moment) set up to run on discover (as we had builds) and points to the 8.9.0 builds made each night with Intel ifort 2021.13. But we can demonstrate it on other machines and in CI.

The bug seems to be triggered by the combination of:

- Intel Fortran compiler 2021.13
- ESMF 8.9 built with ESMF_BOPT=O (optimized)
- Compiling the code with `-fpe0`

With these combinations, things fail.

To run, get a compute node, and then do:
```bash
ml purge
ml comp/gcc/12.3.0 comp/intel/2024.2.0 mpi/impi/2021.13
```
to set up the modules. 

Then we have two makefiles. The first points to the `ESMF_BOPT=g` build and all is well:
```bash
$ make -f GNUmakefile-BOPTg
...
=== Running esmf_ss_reproducer ===
mpirun -np 1 ./esmf_ss_reproducer

=== Running esmf_ss_reproducer_fpe0 ===
mpirun -np 1 ./esmf_ss_reproducer_fpe0
```
All good. Now with `ESMF_BOPT=O`:
```bash
$ make -f GNUmakefile-BOPTO
...
=== Running esmf_ss_reproducer ===
mpirun -np 1 ./esmf_ss_reproducer

=== Running esmf_ss_reproducer_fpe0 ===
mpirun -np 1 ./esmf_ss_reproducer_fpe0
forrtl: error (73): floating divide by zero
Image              PC                Routine            Line        Source
libpthread-2.31.s  00001474F1E6A910  Unknown               Unknown  Unknown
esmf_ss_reproduce  000000000092EA77  Unknown               Unknown  Unknown
libesmf.so         00001474F28EF4A3  _ZN5ESMCI3VMK7sta     Unknown  Unknown
libesmf.so         00001474F28DD349  _ZN5ESMCI2VM7star     Unknown  Unknown
libesmf.so         00001474F25EA091  _ZN5ESMCI6FTable1     Unknown  Unknown
libesmf.so         00001474F25E9CC5  c_esmc_setservice     Unknown  Unknown
libesmf.so         00001474F2ED0685  esmf_gridcompmod_     Unknown  Unknown
esmf_ss_reproduce  0000000000414D8B  esmf_testcase_mod          41  esmf_ss_reproducer.F90
esmf_ss_reproduce  0000000000414D38  esmf_testcase_mod          30  esmf_ss_reproducer.F90
esmf_ss_reproduce  000000000041587F  MAIN__                    132  esmf_ss_reproducer.F90
esmf_ss_reproduce  0000000000414B6D  Unknown               Unknown  Unknown
libc-2.31.so       00001474E623E24D  __libc_start_main     Unknown  Unknown
esmf_ss_reproduce  0000000000414A9A  Unknown               Unknown  Unknown

===================================================================================
=   BAD TERMINATION OF ONE OF YOUR APPLICATION PROCESSES
=   RANK 0 PID 246011 RUNNING AT borgq095
=   KILLED BY SIGNAL: 6 (Aborted)
===================================================================================
make: *** [GNUmakefile-BOPTO:33: run] Error 255
```

theurich:
This PR resolves #464.

Code that utilizes the `log10()` library function is reformulate to ensure  argument is `>0` under _all_ conditions, even those where `log10()` is protected by conditional and should never be called! This is necessary to prevent compiler pre-computation to raise FPE that are trapped when main program is compiled with FPE trapping, i.e. `-fpe0` flag under Intel.

### DIFF
--- a/src/Infrastructure/VM/src/ESMCI_VMKernel.C
+++ b/src/Infrastructure/VM/src/ESMCI_VMKernel.C
@@ -2596,8 +2596,8 @@ void *VMK::startup(class VMKPlan *vmp, void *(fctp)(void *, void *),
     pos = stdTemp.rfind('*');  // right most asterisk
     if (pos != string::npos){
       // found wildcard -> replace with local pet number
-      int digits = 1; // default number of digits needed
-      if (new_npets>1) digits = (int) log10(new_npets-1) + 1;
+      int digits = (int) log10(new_npets);    // always safe
+      if (digits * 10 != new_npets) ++digits; // correct number of digits
       std::stringstream label;                    // fill with zeros from left
       label << setw(digits) << setfill('0') << to_string(sarg[i].mypet);
       sarg[i].stdoutName = stdTemp.substr(0, pos) + label.str()
@@ -2613,8 +2613,8 @@ void *VMK::startup(class VMKPlan *vmp, void *(fctp)(void *, void *),
     pos = stdTemp.rfind('*');  // right most asterisk
     if (pos != string::npos){
       // found wildcard -> replace with local pet number
-      int digits = 1; // default number of digits needed
-      if (new_npets>1) digits = (int) log10(new_npets-1) + 1;
+      int digits = (int) log10(new_npets);    // always safe
+      if (digits * 10 != new_npets) ++digits; // correct number of digits
       std::stringstream label;                    // fill with zeros from left
       label << setw(digits) << setfill('0') << to_string(sarg[i].mypet);
       sarg[i].stderrName = stdTemp.substr(0, pos) + label.str()


danrosen25 (commented):
My only question is what happens when `new_npets=0`. Wouldn't that represent an error? Is there a check that `new_npets` is never 0?

theurich:
> My only question is what happens when `new_npets=0`. Wouldn't that represent an error? Is there a check that `new_npets` is never 0?

`new_npets==0` would mean a VM without PETs, which is not a case supported by ESMF (because I don't know what it would mean).  Execution would never get into `startup()`. I.e. inside of startup() `new_npets>0` is a guaranteed condition for all VMs.

PR #471, opened 2025-07-28
Hot fix for some badly broken code I brought in

theurich:
Fix the "correct the number of digits" code that is badly broken that came in with PR #467.

### DIFF
--- a/src/Infrastructure/VM/src/ESMCI_VMKernel.C
+++ b/src/Infrastructure/VM/src/ESMCI_VMKernel.C
@@ -2597,7 +2597,12 @@ void *VMK::startup(class VMKPlan *vmp, void *(fctp)(void *, void *),
     if (pos != string::npos){
       // found wildcard -> replace with local pet number
       int digits = (int) log10(new_npets);    // always safe
-      if (digits * 10 != new_npets) ++digits; // correct number of digits
+      // correct the number of digits
+      int test_npets = 1;
+      for (int j=0; j<digits; j++)
+        test_npets *= 10;
+      if (test_npets != new_npets) ++digits;
+      else if (test_npets == 1) ++digits;
       std::stringstream label;                    // fill with zeros from left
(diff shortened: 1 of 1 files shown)


src/Infrastructure/VM/src/ESMCI_VMKernel.C
```diff
@@ -2597,7 +2597,12 @@ void *VMK::startup(class VMKPlan *vmp, void *(fctp)(void *, void *),
     if (pos != string::npos){
       // found wildcard -> replace with local pet number
       int digits = (int) log10(new_npets);    // always safe
-      if (digits * 10 != new_npets) ++digits; // correct number of digits
+      // correct the number of digits
+      int test_npets = 1;
+      for (int j=0; j<digits; j++)
+        test_npets *= 10;
+      if (test_npets != new_npets) ++digits;
```
uturuncoglu, line 2604: Maybe you might want to confider to combine these two if blocks since they are doing same thing. Just an idea.

uturuncoglu (approved):
It looks good to me.

danrosen25 (changes requested):
Hi @theurich 
If you're trying to count the number of digits in a integer do this.
```c++
  int digits;
  if (new_npets == 0) {
    digits = 1;
  } else if (new_npets < 0) {
    digits = static_cast<int>(floor(log10(-new_npets))) + 1;
  } else {
    digits = static_cast<int>(floor(log10(new_npets))) + 1;
  }
```

danrosen25:
@theurich 
Yes, but if it's not cleanly readable then it won't be easy to maintain in the future by a future developer. I think volatile is supposed to be marked on the variable on the lhs of an equation within a conditional statement to stop optimization of that condition, meaning it won't try to calculate digits with optimizations.

theurich:
I test with the original code, but `new_npets` declared as `volatile`. It does pass with the reproducer supplied under #464. I agree that this looks like a cleaner solution all around. I am closing this PR in favor of #472.

PR #472, opened 2025-07-29
Alternative approach to prevent log10() pre-computation in optimized mode

theurich:
This PR implements an alternative solution to prevent the pre-computation of  `log10(new_npets-1)` in optimized mode. Thanks to discussion with @danrosen25 under #471, the new approach leverages the `volatile` modifier for targeted prevention of compiler optimization.

I successfully tested this new approach with the reproducer supplied under #464.


oehmke (approved):
That looks goods to me. It's a little more understandable since the explanation and fix are all on the one line. Also, the implementation is what you would expect. Still weird to me that the compiler is breaking this...

theurich:
@mathomp4 just a heads up about this PR. It addresses #464 in a more elegant way. I verified with the reproducer, but want to make sure you are aware of the change.

theurich:
@mathomp4 I just merged this PR into https://github.com/esmf-org/esmf/tree/release/8.9.0. If you have the time to give that branch a quick test, please go ahead. Thank you!

PR #503, opened 2025-11-12
Remove ESMF_ClockGetRingingAlarms from doc

### LINKED ISSUE #469: Documentation bug: ESMF_ClockGetRingingAlarms()
In the description of ESMF_ClockAdvance() there is mention of ESMF_ClockGetRingingAlarms(), but that procedure does not appear to exist.  (At least not in the refdoc.)

"Advances the clock’s current time by one time step: either the clock’s, or the passed-in timeStep (see below).
When the clock is in ESMF_DIRECTION_FORWARD (default), this method adds the timeStep to the
clock’s current time. In ESMF_DIRECTION_REVERSE, timeStep is subtracted from the current time. In either
case, timeStep can be positive or negative. See the "direction" argument in method ESMF_ClockSet().
ESMF_ClockAdvance() optionally returns a list and number of ringing ESMF_Alarms. See also method
**ESMF_ClockGetRingingAlarms()**."

danrosen25:
* Change from ESMF_ClockGetRingingAlarms to ESMF_ClockGetAlarmList

### DIFF
--- a/src/Infrastructure/TimeMgr/interface/ESMF_Clock.F90
+++ b/src/Infrastructure/TimeMgr/interface/ESMF_Clock.F90
@@ -317,7 +317,8 @@
 !     current time.  In either case, {\tt timeStep} can be positive or negative.
 !     See the "direction" argument in method {\tt ESMF\_ClockSet()}.
 !     {\tt ESMF\_ClockAdvance()} optionally returns a list and number of ringing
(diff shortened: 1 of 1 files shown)


danrosen25:
@theurich 
Bob hasn't reviewed. This text makes sense to you, right?
ESMF_ClockGet_doc_update

oehmke:
Sorry, I’ll do it now.

oehmke (approved):
Looks good to me. Thanks for making that change.

Issue #448, opened 2025-07-01
HConfig Issue with Merging using ESMF_HConfigAdd


bena-nasa:
This issue is meant to accompany the corresponding reproducer I will send to @oehmke 
You have two hconfigs from  2 yaml files that define that same key:
In the subsequent examples I will refer to the following 2 yaml files as c1 and c2 respectively

```
Collections:
  c1: {template: foo1}
Exports:
  e1: {collection: c1}
```

and
```
Collections:
  c2: {template: foo2}
Exports:
  e2: {collection: c2}

```

you want to merge them into a single, new hconfig (I will refer to this as merged_config) that will look like this.

```
Collections:
  c1: {template: foo1}
  c2: {template: foo2}
Exports:
  e1: {collection: c1}
  e2: {collection: c2}
```

I will separately send the code used to merge these two. I am unable to merge these two hconfigs together. 

Case 1
hconfigs for c1 and c2 hard coded in source using `ESMF_HConfigCreate(content=...`
merged_config created as an empty map using `ESMF_HConfigCreate(content='{}')`

This produces a file with duplicate keys because after adding a key using EMSF_HConfigAdd, IsDefined says there is not key with that name!
```
Collections:
  c1: {template: foo1}
Exports:
  e1: {coll: c1}
Collections:
  c2: {template: foo2}
Exports:
  e2: {coll: c2}
```

Case2
hconfigs for c1 and c2 hard coded in source using `ESMF_HConfigCreate(content=...`
merged_config with no arguments using `ESMF_HConfigCreate()`

However, to use this, you must not check the return code of `ESMF_HConfigIsDefined`

That manages to merge one key
```
Collections:
  c2: {template: foo2}
  c1: {template: foo1}
Exports: {e1: {coll: c1}}
Exports: {e2: {coll: c2}}
```

bena-nasa:
I'll put the source code here that I'm using to try to merge:
```
#define _VERIFY(A) if (A/=0) then; print*,"Error on line: ", __LINE__; call MPI_Abort(MPI_COMM_WORLD,error_code,status); endif
#define _RC_(rc,status) rc=status);_VERIFY(status
#define _RC _RC_(rc,status)
program test_merge
use ESMF
use mpi
implicit none

integer :: status, error_code
type(ESMF_HConfig) :: h1, h2, merged_hconfig


call ESMF_Initialize(logKindFlag=ESMF_LOGKIND_MULTI, logAppendFlag=.false. )

! case1
[...]

PR #508, opened 2025-12-13
Fix issue in HConfig where checking IsDefined for a non-existent key causes problems with future checks

oehmke:
Fix for issue [#448](https://github.com/esmf-org/esmf/issues/448). In this issue NASA discovered that doing an ESMF_HConfigIsDefined() check with a missing key seems to alter the HConfig object which causes problems for future checks. The yaml-cpp documentation isn't very clear on whether doing [] on a non-existent key modifies the object. However, experimentally it seems to be so. To fix this, I changed the input to Hconfig find() to be const, so that it can't modify it. (This also shouldn't hurt anything since find() shouldn't modify the incoming object.) Doing this now returns an invalid YAML::Node when the object isn't found. This throws an exception with certain operations, so in that case, I now return a valid, but undefined Node. This fixes the NASA issue and passes all of the tests. However, since @theurich is the original developer, he should definitely take a look.

### DIFF
--- a/src/Infrastructure/HConfig/src/ESMCI_HConfig.C
+++ b/src/Infrastructure/HConfig/src/ESMCI_HConfig.C
@@ -640,10 +640,11 @@ HConfig HConfig::createAt(
 //-----------------------------------------------------------------------------
 template <typename T> inline YAML::Node HConfig::find(const T self, HConfig *key){
   if ((*key->doc)[0].IsScalar()){
+    
     // Handle scalar key as simple string direct indexing
     // If key is found by [], it'll return an invalid Node that evaluates as true, so return that.
     // Otherwise, return an undefined Node.
-    if ((YAML::Node tmp = (self)[(*key->doc)[0].as<std::string>()])) return tmp;
+    if (YAML::Node tmp = (self)[(*key->doc)[0].as<std::string>()]) return tmp;
     else return YAML::Node(YAML::NodeType::Undefined);
   }else{
     // handle complex key by iteration and comparing serialized node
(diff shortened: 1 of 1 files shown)


src/Infrastructure/HConfig/src/ESMCI_HConfig.C
```diff
@@ -638,10 +638,14 @@ HConfig HConfig::createAt(
...
 
 //-----------------------------------------------------------------------------
-template <typename T> inline YAML::Node HConfig::find(T self, HConfig *key){
+template <typename T> inline YAML::Node HConfig::find(const T self, HConfig *key){
   if ((*key->doc)[0].IsScalar()){
-    // handle scalar key as simple string direct indexing
-    return (self)[(*key->doc)[0].as<std::string>()];
+    
+    // Handle scalar key as simple string direct indexing
+    // If key is found by [], it'll return an invalid Node that evaluates as true, so return that.
```
theurich:
```suggestion
    // Handle scalar key as simple string direct indexing
    // If key is found by [], it'll return an invalid Node that evaluates as true,
    // so return that.
```

theurich: Looks good. Thanks for fixing this!

Only small suggestion to keep with (my typical) 80 char line limit. :-)

theurich (changes requested):

oehmke:
@theurich, thanks for reviewing!  I agree about the 80 character limit, but sometimes my terminals creep wider and I don't notice. :-) I accepted that change, so we should be good for that too.

theurich:
> @theurich, thanks for reviewing! I agree about the 80 character limit, but sometimes my terminals creep wider and I don't notice. :-) I accepted that change, so we should be good for that too.

Thanks! Nice to have the issue fixed!

PR #533, opened 2026-04-07
Simplify around ESMX custom CMAKE Fortran build flags

theurich:
This PR simplifies the way that ESMX sets custom `CMAKE_Fortran_FLAGS*` variables. Only set these flags in those cases where we want to use specific non-default options. Use default options as much as possible, and don't warn about that (since it is best practice). Add custom debug flags for `LLVMFlang`, i.e. the "new" `flang` implementation that now comes with LLVM.


theurich:
Also see fix e7b9c3c4de26bb4b3b59d38110ec68f67cde7d79

PR #545, opened 2026-04-23
Output esmpy json report to named file

danrosen25:
This will help preserve the json summary report from pytest, which is processed by esmf-test-scripts. The report is bootstrapped with a comment, so if it crashes then the file will still exist.

Once merged the `esmpy${VERSION}-petx${NP}-summary.json` should be added to the esmf-test-scripts artifacts and used when creating the summary.


theurich (approved):
Looks good to me.

PR #547, opened 2026-05-12
Bring ESMX_Data enhancements back into ESMF

### LINKED ISSUE #521: Merge ESMX_Data component enhancements into the official ESMF develop branch
A number of enhancements for ESMX_Data were implemented as part of the Navy-ESPC work. These changes need to be merged into the mainline ESMF development branch.

Also must consider PR #483 that also brings in changes to ESMX_Data.

### LINKED ISSUE #551: ESMX_Data improvements after UI review
We had a review of the ESMX_Data user interface (UI) during the 05/27/26 core team meeting. We identified several areas of improvement. This serves as an umbrella issue, with specific improvement requests handled as sub issues.

### LINKED ISSUE #552: ESMX_Data - output handling
The decision was made to handle output of field data to file under ESMX_Data separate from the `Diagnostic` attribute. Instead a new YAML map with key `outputs` will be introduced that for now holds a single map with key `fieldList`. The value of that map is a list of field names.

```
  outputs:
      fieldList:
        - field1
        - field2
        - field3
```

Fields listed are going to be written out using the default NUOPC output procedure (i.e. NetCDF format and default file names).

The idea is that `outputs` in the future can be extended to be a list of maps, for different output streams, and each map potentially holding more than just the `fieldList` key, allowing for additional output specification such as file format, file naming, output frequency.

### LINKED ISSUE #553: ESMX_Data - dataValidate changes
The decision was made to change `dataValidate` in the following way:

```
dataValidate:
  min: <value>      # unchanged
  max: <value>      # unchanged
  mask: <value>     # unchanged
  print: on|off     # changed from "diagnose" for clarity that this is about printing info to stdout
  action:  ignore|warning|error         # with "error" as default. 
```

### LINKED ISSUE #554: ESMX_Data - staggerLoc complexity
In order to support all of the ESMF regridding methods, Grids used to create Fields must sometimes provide more than one staggerLoc. In order to support this, the `geometries` definition needs to support a _list_ option for `staggerLoc`.

```
geometries:

    globalSingleStagger:
      geom:           grid1PeriDimUfrm
      minCornerCoord: [-180, -89]
      maxCornerCoord: [+180, +89]
      maxIndex:       [ 100,  50]
      staggerLoc:     center               # <-- single stagger option

    globalMultiStagger:
      geom:           grid1PeriDimUfrm
      minCornerCoord: [-180, -89]
      maxCornerCoord: [+180, +89]
      maxIndex:       [ 100,  50]
      staggerLoc:     [center, corner]   # <-- multi stagger option, e.g. to support conservative regridding
```

A field can then be created as before when referencing a single stagger geom:

```
  exportFields:
    sea_surface_temperature:
      geometry:     globalSingleStagger   #  <-- scalar string value
```
Or must specify the staggerLoc explicitly when referencing a multi stagger geom:
```
  exportFields:
    sea_surface_temperature:
      geometry:     globalMultiStagger : center  #  <-- geom name is key with staggerLoc (or meshloc when we support Mesh) as the value
```

### LINKED ISSUE #555: ESMX_Data - dataInit dependencies
We want ESMX_Data to implement fully compliant interaction with the NUOPC data-dependencies-during-initialization protocol. For this I propose the following refined ESMX_Data behavior as it relates to the `dataInit` option for fields in `importFields` and `exportFields`:

- `dataInit` for a field in `exportFields`:
  - If absent: does not initialize the field during DataInitialize and do not mark it as NUOPC `Updated`. This will trigger a NUOPC compliance error as per protocol if any other component depends on this field during initialize.
  - If present: initializes the field during DataInitialize according to the `dataInit` dynamic-arithmetic-expression. If variables from the `importFields` are involved, only initialize when all of those import fields are at the current timestamp. Only then mark the initialized export field as NUOPC `Updated`.

- `dataInit` for a field in `importFields`:
  - If absent: only marks as NUOPC `Updated` once the timestamp has been updated to the current time via a Connector, i.e. this triggers the data-dependency-during-initialize protocol.
  - If present: initializes the field during DataInitialize according to the `dataInit` dynamic-arithmetic-expression. It is okay for this expression to include other import fields. Only initialize when all of those import fields are at the current timestamp. Only then mark the initialized import field as NUOPC `Updated`.

The above allows an ESMX_Data instance to selectively trigger the data-dependency-during-initialization protocol for specific fields in its importState, or alternatively locally initialize them.

theurich:
This PR is close to a full rewrite of the `ESMX_Data` implementation. The bulk of this implementation was developed under the NRL NEPTUNE and ESPC projects, where we use ESMX to build, run, and validate components and applications. Key aspects of this work include:
- Support for multiple geom definitions
- Each field to specify the associated geom, can be different for different fields
- Grid definitions to support periodic options (for global grids), as well as support local area grids
- Better diagnostic output options for fields, with options to diagnose, issue warnings or errors
- Use of the standard NUOPC `Diagnostic` attribute to control field data output to NetCDF.
- More generic options for field data initialization and advance, with support for use of import field data in simple arithmetic expressions for feedback
- Support switching between Mediator and Model timekeeping mode. This affects time stamping of export fields. It affects where in the run sequence the ESMX_Data instance can/should be placed.

This PR supersedes #483, and closes #521.

The [README.md](https://github.com/esmf-org/esmf/tree/feature/esmx-data/src/addon/ESMX/Comps/ESMX_Data#readme) associated with this PR provides a useful high-level overview of the features provided by the proposed `ESMX_Data` implementation.

The associated nuopc-app-prototypes PR is https://github.com/esmf-org/nuopc-app-prototypes/pull/16

---

See https://github.com/esmf-org/esmf/issues/551 for UI enhancements identified during review with core team. This PR resolves #551.
- Resolves #552
- Resolves #553 
- Resolves #554
- Resolves #555


theurich:
See #551 for UI enhancements identified during review with core team.

danrosen25:
### General Documentation Notes (Questions and Comments)
esmf version: dbfe20c prototypes version: [0b9ca4c](https://github.com/esmf-org/nuopc-app-prototypes/commit/0b9ca4ccc360dc5b0b9f846c2c4e961c656816bc) 

@theurich The following is a list of thoughts I had running ESMX tests, reading the ESMX documentation, and reading the ESMX_Data documentation. Code review to follow.

1. In order to build ESMX prototypes ESMF_ROOT or CMAKE_PREFIX_PATH must point to an ESMF installation containing ESMFConfig.cmake.  ESMF_ROOT is automatically set by install-esmf-action (similar to ESMFMKFILE) but ./testProtos.sh will fail on other systems.
2. What happens if multiple fields with the same name are added to an output bundle?
3. README "CAUTION" text boxes are more visible using [GitHub alerts MD extension](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)
4. Typo "esmfRun.yaml" is written instead of "esmxRun.yaml" in [ESMX README](https://github.com/esmf-org/esmf/blob/dbfe20cccf4031fb0085a04e36c5c479ee40c69d/src/addon/ESMX/README.md?plain=1#L334)
5. Is this sentence correct because it sounds model: ESMX_Data is not the Component Label? ["All standard ESMX Component Label Options are supported, in particular model: ESMX_Data"](https://github.com/esmf-org/esmf/blob/dbfe20cccf4031fb0085a04e36c5c479ee40c69d/src/addon/ESMX/Comps/ESMX_Data/README.md?plain=1#L29)
6. In Data Validation - Do you think maskVal might make more sense? The user is providing a maskVal and a mask is being computed based on this? How about minVal and maxVal to match?
7. In Dynamic Arithmetic Expressions what happens when the input field array shape doesn't match the current field? It's not documented but I assume it fails.
8. In Dynamic Arithmetic Expressions I'd prefer a more common special character prefix for input fields and special context variables, $ being the standard
9. In 9.0.0 do we want to remove the ESMX testing infrastructure (currently says beta). I think there could be a better framework for testing that exists as a tool outside of ESMX.

theurich:
> ### General Documentation Notes (Questions and Comments)
> esmf version: [dbfe20c](https://github.com/esmf-org/esmf/commit/dbfe20cccf4031fb0085a04e36c5c479ee40c69d) prototypes version: [0b9ca4c](https://github.com/esmf-org/nuopc-app-prototypes/commit/0b9ca4ccc360dc5b0b9f846c2c4e961c656816bc)
> 
> @theurich The following is a list of thoughts I had running ESMX tests, reading the ESMX documentation, and reading the ESMX_Data documentation. Code review to follow.
> 
> 1. In order to build ESMX prototypes ESMF_ROOT or CMAKE_PREFIX_PATH must point to an ESMF installation containing ESMFConfig.cmake.  ESMF_ROOT is automatically set by install-esmf-action (similar to ESMFMKFILE) but ./testProtos.sh will fail on other systems.

Associated issue https://github.com/esmf-org/nuopc-app-prototypes/issues/21, and handled by https://github.com/esmf-org/nuopc-app-prototypes/pull/22.

> 2. What happens if multiple fields with the same name are added to an output bundle?

I will look into that and report back.

> 5. Is this sentence correct because it sounds model: ESMX_Data is not the Component Label? ["All standard ESMX Component Label Options are supported, in particular model: ESMX_Data"](https://github.com/esmf-org/esmf/blob/dbfe20cccf4031fb0085a04e36c5c479ee40c69d/src/addon/ESMX/Comps/ESMX_Data/README.md?plain=1#L29)

That is right, `model: ESMX_Data` is _not_ the Component Label. `MyDat:` or what ever key is used to introduce a specific component instance is the "Component Label", and `model:` is one of the "standard ESMX Component Label Options". 

> 6. In Data Validation - Do you think maskVal might make more sense? The user is providing a maskVal and a mask is being computed based on this? How about minVal and maxVal to match?

I assume `Val` would stand for _value_? If so, I would suggest spelling it out, especially because it could be confused standing for _validation_. So it would be `minValue`, `maxValue`, and `maskValue`. I see what you mean with the value of the mask not being the same as the mask. However, I don't expect there would be much practical confusion, and `min`, `max`, `mask` are just shorter.  I am fine either way. Let's see what other's think on the core call.

> 7. In Dynamic Arithmetic Expressions what happens when the input field array shape doesn't match the current field? It's not documented but I assume it fails.

It's not a supported condition, but I cannot remember whether I have an ERROR check for that or not. I will look into it and report back.

> 8. In Dynamic Arithmetic Expressions I'd prefer a more common special character prefix for input fields and special context variables, $ being the standard

There is no technical need to add a special character prefix for input fields. It's basically just like in the BASH `(( ... ))` arithmetic expressions, where variables also do not need (and better should not have) the `$` prefix. Clean arithmetic expressions using the plain input field names, without special prefixing.

The `$` can become especially problematic in the `esmxRun.yaml` when also doing some pre-processing with variable replacement. E.g. we do this for NEPTUNE, where `startTime` and `stopTime` are set in `esmxRun.yaml` via pre-processing, using `$` syntax. This is a pretty "dumb" preprocessor that would likely trip over `$` anywhere else in the file.

Of course the special context variables need to be distinguished from input fields, and use the underscore "_" prefix. This is similar to the common Python convention of using underscore prefix for internal context variables or helper functions.

> 9. In 9.0.0 do we want to remove the ESMX testing infrastructure (currently says beta). I think there could be a better framework for testing that exists as a tool outside of ESMX.

I have not used the ESMX Testing Infrastructure myself. Since it has been saying "beta" all along, I see no problem removing it for 9.0.0 for a cleaner path forward.

theurich:
> 3. README "CAUTION" text boxes are more visible using [GitHub alerts MD extension](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts)

See https://github.com/esmf-org/esmf/pull/547/commits/4659727a82ccd9ca7843aae98db53af6f22d35c2

> 4. Typo "esmfRun.yaml" is written instead of "esmxRun.yaml" in [ESMX README](https://github.com/esmf-org/esmf/blob/dbfe20cccf4031fb0085a04e36c5c479ee40c69d/src/addon/ESMX/README.md?plain=1#L334)

Fixed in https://github.com/esmf-org/esmf/pull/547/commits/c295be96a4f1dcd126873e0d740c2c4f7f446052

theurich:
> > 9. In 9.0.0 do we want to remove the ESMX testing infrastructure (currently says beta). I think there could be a better framework for testing that exists as a tool outside of ESMX.
> 
> I have not used the ESMX Testing Infrastructure myself. Since it has been saying "beta" all along, I see no problem removing it for 9.0.0 for a cleaner path forward.

Opened #582 to track removal of ESMX Testing Infrastructure for 9.0.0.

theurich:
> > 5. Is this sentence correct because it sounds model: ESMX_Data is not the Component Label? ["All standard ESMX Component Label Options are supported, in particular model: ESMX_Data"](https://github.com/esmf-org/esmf/blob/dbfe20cccf4031fb0085a04e36c5c479ee40c69d/src/addon/ESMX/Comps/ESMX_Data/README.md?plain=1#L29)
> 
> That is right, `model: ESMX_Data` is _not_ the Component Label. `MyDat:` or what ever key is used to introduce a specific component instance is the "Component Label", and `model:` is one of the "standard ESMX Component Label Options".

See 85dd28f for reworked wording.

> > 6. In Data Validation - Do you think maskVal might make more sense? The user is providing a maskVal and a mask is being computed based on this? How about minVal and maxVal to match?
> 
> I assume `Val` would stand for _value_? If so, I would suggest spelling it out, especially because it could be confused standing for _validation_. So it would be `minValue`, `maxValue`, and `maskValue`. I see what you mean with the value of the mask not being the same as the mask. However, I don't expect there would be much practical confusion, and `min`, `max`, `mask` are just shorter. I am fine either way. Let's see what other's think on the core call.

During the core call it was decided to go with `min`, `max`, `skip`. 1ec5839 implements this change, together with support for scalar or list valued 'skip' specification and generalization of the supporting FieldStats() implementation.

theurich:
> > 2. What happens if multiple fields with the same name are added to an output bundle?
> 
> I will look into that and report back.

I checked for both `importFields` and `exportFields`. Both cases trigger the expected ERROR and abort:

```
20260806 083237.377 ERROR            PET0 ESMF_StateInternals.F90:1083 ESMF_StateClsAddList       Arguments are incompatible  - existing item sea_surface_temperature not replacable
```

theurich:
> > 7. In Dynamic Arithmetic Expressions what happens when the input field array shape doesn't match the current field? It's not documented but I assume it fails.
> 
> It's not a supported condition, but I cannot remember whether I have an ERROR check for that or not. I will look into it and report back.

I tested, and there was no explicit ERROR check, so in my setup with Fortran bounds checking turned on, it would crash with a Fortran runtime error as expected at this statement:
```
stack(:,top) = fPtrImportR8
```
I implemented explicit ERROR checking with e6b0cf7, so now you'd get a clean abort with ERROR:
```
Arguments are incompatible  - Import fields in arithmetic expression 'sea_surface_temperature + sea_surface_height_above_sea_level' must match size of targeted export field!
```

theurich:
> > 8. In Dynamic Arithmetic Expressions I'd prefer a more common special character prefix for input fields and special context variables, $ being the standard
> 
> There is no technical need to add a special character prefix for input fields. It's basically just like in the BASH `(( ... ))` arithmetic expressions, where variables also do not need (and better should not have) the `$` prefix. Clean arithmetic expressions using the plain input field names, without special prefixing.
> 
> The `$` can become especially problematic in the `esmxRun.yaml` when also doing some pre-processing with variable replacement. E.g. we do this for NEPTUNE, where `startTime` and `stopTime` are set in `esmxRun.yaml` via pre-processing, using `$` syntax. This is a pretty "dumb" preprocessor that would likely trip over `$` anywhere else in the file.
> 
> Of course the special context variables need to be distinguished from input fields, and use the underscore "_" prefix. This is similar to the common Python convention of using underscore prefix for internal context variables or helper functions.

As decided during the core meeting 08/05/26, I implemented support to allow import fields with names starting with underscore `_`. At the same time I implemented error checking, at the point of usage in arithmetic expressions, where such field name might conflict with the special context variables. See 71f572b for the implementation, and https://github.com/esmf-org/esmf/tree/feature/esmx-data/src/addon/ESMX/Comps/ESMX_Data#3-operands-and-variables for the relevant updated documentation.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
```diff
@@ -1,1648 +1,2502 @@
...
-          line=__LINE__, file=__FILE__)) return
-        call ESMF_HConfigDestroy(xdatacfg, rc=rc)
+          line=__LINE__, file=__FILE__)) return  ! bail out
+        GeomCreateFromHConfig = ESMF_GeomCreate(grid, rc=rc)
         if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, &
-          line=__LINE__, file=__FILE__)) return
-      endif ! xdatacfg
-    endif ! config
+          line=__LINE__, file=__FILE__)) return  ! bail out
+!      else if (geom == "mesh") then
```
danrosen25, line 1121: Remove unused code.
```suggestion
```

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
```diff
@@ -1,1648 +1,2502 @@
...
-    call ESMF_GridCompGetInternalState(xdata, is, rc)
-    if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, &
-      line=__LINE__, file=__FILE__)) return
-    xstate => is%ptr
-    if (.not. associated(xstate)) then
-      call ESMF_LogSetError(ESMF_RC_PTR_NOTALLOC, &
-        msg='XDATA: xstate has not been associated', &
+    if (timeKeeping /= "MODEL" .and. timeKeeping /= "MEDIATOR") then
+      call ESMF_LogSetError(ESMF_RC_ARG_WRONG, &
+        msg="An invalid setting was found for 'timeKeeping' "// &
```
danrosen25, line 232: Suggestion: Add a note suggesting the user see the ESMX documentation OR add the valid values of Model or Mediator to this error message.

src/addon/ESMX/Comps/ESMX_Data/ESMX_DataProcess.F90
```diff
@@ -0,0 +1,995 @@
+module dataProcess
```
danrosen25: Prepend namespace for file and module. Instead of `dataProcess` call it `ESMX_DataProcess`. This will help for symbol and modfile collisions and it matches the filename.

src/addon/ESMX/Comps/ESMX_Data/ESMX_DataProcess.F90
```diff
@@ -0,0 +1,995 @@
...
+
+  !-----------------------------------------------------------------------------
+
+  contains
+
+  !-----------------------------------------------------------------------------
+
+  subroutine process(importState, expression, exportField, step, okayTime, &
+    importsOkay, rc)
+    ! Process according to the expression infix string and store in exportField
```
danrosen25: Subroutine arguments aren't documented

src/addon/ESMX/Comps/ESMX_Data/dataProcess.F90
```diff
@@ -0,0 +1,995 @@
...
+        line=__LINE__, file=__FILE__, rcToReturn=rc)
+      return ! bail out
+    endif
+
+  end subroutine
+
+  !-----------------------------------------------------------------------------
+
+  subroutine normalize_infix(input, output)
+    ! Normalize the incoming infix string with single white space deliminators
```
danrosen25:
```suggestion
    ! Normalize the incoming infix string with single white space delimiters
```

src/addon/ESMX/Comps/ESMX_Data/ESMX_DataProcess.F90
```diff
@@ -0,0 +1,995 @@
...
+    needs_leading_zero = .true.
+    in_operand = .false.
+
+    do i = 1, len_trim(input)
+      c = input(i:i)  ! current character
+
+      if (c == " ") then
+        if (in_operand) then
+          ! finish operand by adding trailing space
+          output = output // " "
```
danrosen25, line 516: Note: This type of concatenation is slow because it's reallocating the array each time a new character is appended. I think it's okay for now but might be something we want to enhance if we want this to be a well utilized feature in ESMF.

src/addon/ESMX/Comps/ESMX_Data/ESMX_DataProcess.F90
```diff
@@ -0,0 +1,995 @@
...
+    logical                         :: needs_leading_zero, in_operand
+
+    output = ""
+    needs_leading_zero = .true.
+    in_operand = .false.
+
+    do i = 1, len_trim(input)
+      c = input(i:i)  ! current character
+
+      if (c == " ") then
```
danrosen25, line 513: Note: We're only treating ` ` as whitespace but there are other types of whitespace, such as tabs.

src/addon/ESMX/Comps/ESMX_Data/dataProcess.F90
```diff
@@ -0,0 +1,995 @@
...
+        cycle
+      end if
+
+      if (is_boundary(c)) then
+        if (in_operand) then
+          ! finish operand by adding trailing space
+          output = output // " "
+          in_operand = .false.
+        end if
+        ! Handle Unary Plus and Minus
```
danrosen25: Trying to wrap my head around this but `3*-2` would normalize to `3 * 0 - 2`, isn't that a problem? I'll look through the rpn code but I assume this would evaluate to `3 0 * 2 - = -2` instead of -6.

danrosen25:
@theurich
### Issue 1:
I just wanted to give you a heads up. normalize_infix is broken due to injecting of `"0 " // operator` token, it will cause an incorrect answer. For example `3 * -2 = -6` but this will normalize to `3 * 0 - 2`. The expression converts to rpn `3 0 * 2 - = -2`. `-2`. It needs parentheses to solve the inverse calculation first `3 * ( 0 - 2 )` converts to 3 0 2 - *. I've been trying to solve this for you but there are many cases to consider, such as `-(3 * 2)`, which should evaluate to `0 3 2 * -`. A better solution may exist in the rpn shunting yard algorithm but I've been trying to fix it in the normalize_infix without success.
### Issue 2:
We need to check for unmatched parentheses in the rpn subroutine.  An expression like `3 * ((2 - 1) - 3 * 2` needs to return an error. This very likely could be a typo and where the closing parenthesis lands changes the value.

theurich:
@danrosen25 I just pushed a bunch of updates following your last comment. I went through several iterations, and also noticed a bunch of other "edge cases" that needed fixing (big one was that scientific notation like 1e-10, where there is an explicit sign in the exponent, broke things). Plus there was no support for power, which I came across when I was thinking about stacking operators more generally like `**`. I ended up with `^` for power, which is now implemented and documented.

As for the main issue of handling unary `+` and `-`, and stacking of them, thank you for the `infix_to_rpn.F90` code you sent. I think the idea of making `infix_to_rpn()` a stateful method as it processes the input makes perfect sense, and since this is on a level where we write directly into the RPN stream (no more parenthesis issues), it makes the whole inserting of zero business so much easier. Inspired by your code but still wanting to stay with the tokenized approach (which I think provides for better contextual error handling), I move the unary handling into `infix_to_rpn()`. 

I think the current version should be pretty robust now. I tested with a bunch of different cases. I also added more error handling. Unbalanced parentheses are now flagged. Plus invalid unary or adjacent operators are detected and error out.

danrosen25:
@theurich 
I ran tests on `normalize_infix` and `infix_to_rpn` and found a failure in `infix_to_rpn`.
```
 FAIL: Unary minus at start
   Input:    -3 * 4
   Expected: 0 3 - 4 *
   Got:      0 3 4 * -
```

theurich:
> @theurich I ran tests on `normalize_infix` and `infix_to_rpn` and found a failure in `infix_to_rpn`.
> 
> ```
>  FAIL: Unary minus at start
>    Input:    -3 * 4
>    Expected: 0 3 - 4 *
>    Got:      0 3 4 * -
> ```

I think it is okay for the `*` to carry out before the `-`. I cannot think of a case where that would cause problems - can you? I also just tested `-3 * -4` to make sure:
```
 pre-normalize infix: -3 * -4
 normalized infix: - 3 * - 4
 rpn_expression: 0 3 0 4 - * -
```
Which also looks okay to me.

theurich:
@danrosen25 I am thinking of renaming `dataProcess.F90` -> `ESMX_DataProcess.F90` under the `src/addon/ESMX/Comps/ESMX_Data` directory. I know you had mentioned something similar before. Any thoughts?

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
```diff
@@ -1,1648 +1,2502 @@
...
   !-----------------------------------------------------------------------------
 
   use ESMF
   use NUOPC
-  use NUOPC_Model, &
-    modelSS    => SetServices
+  use NUOPC_ModelBase, &
+    modelBaseSS    => SetServices
+
+  use dataProcess
```
danrosen25: File name was changed to ESMX_DataProcess and I think the module name should be changed as well.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
```diff
@@ -1,1648 +1,2502 @@
...
 
   implicit none
 
   private
 
   public SetServices, SetVM
```
danrosen25, line 18: Where is the SetVM subroutine?

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
```diff
@@ -1,1648 +1,2502 @@
...
     character(:), allocatable  :: badKey
     logical                    :: isFlag
     character(:), allocatable  :: configKey(:)
 
     rc = ESMF_SUCCESS
 
+    ! query the component for info
+    call NUOPC_CompGet(xdata, name=name, rc=rc)
+    if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, &
+      line=__LINE__, file=trim(name)//":"//__FILE__)) return  ! bail out
```
danrosen25: I think we should remove `! bail out` for code readability. It adds nothing.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
```diff
@@ -1,1648 +1,2502 @@
...
+                  outputs(item)%name//"' "// "(maybe a typo?): "//badKey, &
+                line=__LINE__, file=__FILE__, rcToReturn=rc)
+              return  ! bail out
+            endif
+            ! ingest onDataInit
+            outputs(item)%onDataInit = ESMF_HConfigAsLogical(hconfigMap, &
+               keyString="onDataInit", asOkay=isFlag, rc=rc)
+            if (ESMF_LogFoundError(rcToCheck=rc, msg=ESMF_LOGERR_PASSTHRU, &
+              line=__LINE__, file=__FILE__)) return  ! bail out
+            if (.not.isFlag) outputs(item)%onDataInit = .false. ! default
```
danrosen25, line 314: Same as the bail out comments I think `! default` can be removed for code cleanliness.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
```diff
@@ -1,1648 +1,2502 @@
...
+            filestatus=ESMF_FILESTATUS_OLD
+            if (output%onDataInit) then
+              step = stepCounter+1
+            else
+              step = stepCounter
+              if (stepCounter==1) filestatus=ESMF_FILESTATUS_REPLACE
+            endif
+            fileNamePrefix="data_"//trim(name)//"_"//output%name
+            if (output%separateTimeFiles) then
+              write(stepString, "(I4.4)") stepCounter
```
danrosen25: Minor Bug: This means the maximum time step for output is 9999 but there is not guard so output past 9999 will look like ****.

danrosen25 (changes requested):
It's a lot of code to review and this would have been easier if it was submitted as separate PRs. I think that some of the blocks of code can be refactored into subroutines to enhance readability and therefore maintainability but that won't affect the operation and I'd like to move this along sooner. The **only change** I'm really pushing is to change the module name for dataProcess to ESMX_DataProcess. I think if the CMake build system has two components with the same module names and they happen to have the same procedure names it will fail during linking.

Note that a few of my comments have already been fixed.

Overall this is good work and I even learned a few new coding techniques.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
theurich: This is a reminder of where in the future Mesh related code would go. I added `!TODO` note for clarity. See 2c968a3.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
theurich: This is the standard formulation of the ERROR message that is thrown when an invalid setting is found. There are a good number of these error checks, many of them with too many valid options to make it practical to list them all in the error message. I think it is understood that when you get an ERROR about an invalid option you should consult the documentation, not much value in adding that imo.

src/addon/ESMX/Comps/ESMX_Data/ESMX_DataProcess.F90
theurich: I agree. See 8a9498c.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
theurich: I agree. See 8a9498c.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
theurich: The standard `SetVM` is inherited from `NUOPC_ModelBase`.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
theurich: I agree. See 73387b4.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
theurich: I like those `! default` comments in code I write. It helps me identify places where I follow the same coding idiom of setting some default, that might be overridden in the next few lines under certain conditions. There aren't very many of those places, and I find it useful when coming back to this code.

src/addon/ESMX/Comps/ESMX_Data/ESMX_Data.F90
theurich: See dbd63ed. The ultimate limit is now 32 characters, and there is an ERROR check when overflowing that limit. The starting number of digits (with leading 0) is still 4, and will automatically increase as `stepCounter` goes above that, until hitting the 32 digit upper limit.

src/addon/ESMX/Comps/ESMX_Data/ESMX_DataProcess.F90
theurich: See 1c4164f for argument doc.

theurich:
@danrosen25 see 8a9498c for the Fortran module renaming to `ESMX_DataProcess` for consistency with file name.

See your individual suggestions for my responses, and link to code change hash where applicable. I appreciate the thorough review.

danrosen25 (approved):

PR #549, opened 2026-05-22
Generate standard CMake package configuration for ESMF

theurich:
This PR implements generation of a standard CMake package configuration for ESMF during "make install". It does so by ingesting the installed `esmf.mk` file and gleaning the necessary ESMF installation information. The `ESMF_COMM` setting is considered to correctly enable/disable MPI dependency.


theurich:
This is in part responding to a request under Spack: https://github.com/spack/spack-packages/issues/5045

danrosen25:
@theurich 
Do you plan to keep `FindESMF.cmake` and continue to install it even though the `ESMFConfig.cmake` with `ESMF_ROOT` is preferred.? The entire `cmake` directory is being installed and that includes the `ESMFConfig.cmake.in` and `generate_config.py`. If you're intending to keep and install all the files in the `cmake` directory then we should update the `README.md` file that directory.

This also needs some documentation in the Users Guide.

theurich:
> @theurich Do you plan to keep `FindESMF.cmake` and continue to install it even though the `ESMFConfig.cmake` with `ESMF_ROOT` is preferred.? The entire `cmake` directory is being installed and that includes the `ESMFConfig.cmake.in` and `generate_config.py`. If you're intending to keep and install all the files in the `cmake` directory then we should update the `README.md` file that directory.
> 
> This also needs some documentation in the Users Guide.

I was thinking to keep installing `FindESMF.cmake` into `ESMF_INSTALL_CMAKEDIR_ABSPATH`, at least for v9. It is still a valid way of integrating ESMF into a users CMake based build, and provides a bit of backward compatibility for users that were using it.  Only `FindESMF.cmake`  will be copied though, not the entire `cmake` directory. I made that a little more explicit and visible in the `install` target in https://github.com/esmf-org/esmf/pull/549/commits/4ab747f1012c0cc3a41103b4de5cd2162bd14f98, although the previous `*.cmake` pattern was sufficient (but less clear).

As for the `cmake/README.md`, I agree that we should extend this to cover the preferred `ESMFConfig.cmake` approach.

theurich:
This has been successfully tested with the "ESMF Spack Test" workflow: https://github.com/esmf-org/esmf/actions/runs/27219678683 using the following settings:
-    Use workflow from:  `feature/generate-cmake-config`
-    Compiler version:   `gcc@latest`
-    ESMF version:       `esmf@git.4ab747f1012c0cc3a41103b4de5cd2162bd14f98=9.0.0`
-    NUOPC app version:  `feature/generate-cmake-config`

theurich:
@esmf-org/esmf-reviewers I plan to merge this PR, and the associated https://github.com/esmf-org/nuopc-app-prototypes/pull/18 today. Let me know if there are any concerns. Thanks.

PR #565, opened 2026-06-25
Public abstract interfaces for GridComp and CplComp callback routines

### LINKED ISSUE #564: Provide explicit abstract interfaces for user-supplied callback routines
The ESMF component classes implement inlined interfaces for user-supplied dummy procedures. There are advantages of having those interfaces explicitly declared in the ESMF API as `abstract interface`.

Currently MAPL and NUOPC define their own abstract interfaces for this purpose, but this could be cleaned up if ESMF provided them instead.

See also #563 for context.

theurich:
This PR implements abstract interfaces for user-provided user callback routines for `ESMF_GridComp`:
```
  public ESMF_I_GridCompEntryPoint
  public ESMF_I_GridCompServices
  public ESMF_I_GridCompVM
```
and `ESMF_CplComp`:
```
  public ESMF_I_CplCompEntryPoint
  public ESMF_I_CplCompServices
  public ESMF_I_CplCompVM
```
Further leverage the new ESMF provided abstract interfaces on the NUOPC and ESMX layers, removing redundant interface definitions.

Resolves #564.

See branch documentation for review: https://earthsystemmodeling.org/docs/nightly/feature/abstract-interfaces/ESMF_refdoc/node4.html#SECTION04034000000000000000

### DIFF
--- a/.github/workflows/test-coverage.yml
+++ b/.github/workflows/test-coverage.yml
@@ -23,16 +23,19 @@ jobs:
         cd ${{ github.workspace }}/esmf-containers/test-coverage
         docker build . --tag esmf/test-coverage --build-arg ESMF_BRANCH=$GITHUB_REF_NAME
 
-    - name: Copy artifacts
+    - name: Copy and extract artifacts
       run: |
         mkdir -p ${{ github.workspace }}/artifacts
         CID=$(docker run -dit --name runner esmf/test-coverage)
         docker cp ${CID}:/artifacts/test_coverage-artifacts.zip ${{ github.workspace }}/artifacts
         docker stop ${CID}
         docker rm ${CID}
+        # Extract the contents and delete the original zip file
+        unzip ${{ github.workspace }}/artifacts/test_coverage-artifacts.zip -d ${{ github.workspace }}/artifacts
+        rm ${{ github.workspace }}/artifacts/test_coverage-artifacts.zip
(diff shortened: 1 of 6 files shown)


theurich:
@oehmke and @tclune this PR is ready to go into develop. The CI tests are clean, and I also ran the branch on a NAG platform just to be sure. Plus I ran nuopc-app-protos CI test against it. Also clean.

As for the nomenclature of public abstract interfaces provided by ESMF, I decided on the `ESMF_I_` prefix. Since it is part of the ESMF API, it should start with `ESMF_`. Using `I_` for interfaces is common practice, and also shows in the MAPL example Tom provided, so the combination then is `ESMF_I_` prefix. It's also great that this does not cost us many extra characters, which is important in the context of long ESMF names. 

Please let me know what you think. I hope to move this forward quickly. Thanks.

oehmke (approved):
Looks good to me! I'm excited about this change because I feel like it makes things much cleaner to have the interfaces represented this way for the users.

tclune:
Looks good, but what about the interface for userRoutine for ESMF_GridCompSetServices?  

Thanks.

theurich:
> Looks good, but what about the interface for userRoutine for ESMF_GridCompSetServices?
> 
> Thanks.

That's `ESMF_I_GridCompServices`.

PR #566, opened 2026-07-01
Fix the InternalState API and API doc

theurich:
This PR fixes the placement of the `keywordEnforcer` in the InternalState API. It now follows the general rule that `keywordEnforcer` should separate the optional arguments from the leading non-optional arguments. Also ensuring that the API and API doc are consistent.

This resolves https://github.com/esmf-org/esmf-support/issues/604 

For updated API doc review see: https://earthsystemmodeling.org/docs/nightly/fix/internal-state-api/ESMF_refdoc/node4.html#SECTION04084000000000000000


oehmke (approved):
Looks good! Thanks for fixing that.

PR #523, opened 2026-02-17
ESMC_DynamicMask with ESMPy access

theurich:
This PR is based on the work of @bena-nasa that he provided through #365, where he implemented predefined DynamicMask routines, added the ESMC_DynamicMask API, routed it through to the ESMC_Field API and to ESMPy. 

This PR supersedes #365. It has been updated with changes from `develop` up to beta tag `v9.0.0b10`, and includes a number of fixes, nomenclature cleanups, and formatting changes.

There are currently a few outstanding issues. Some are reflected by failing CI tests: https://github.com/esmf-org/esmf/actions/runs/22085849227. There are a number of areas that need attention before this PR is ready for merging:

1. The `ESMF_FieldRegridDynMaskPredef*UTest.F90` tests need to be reviewed and completed/corrected. Specifically the `V` and `ungridded` sub-tests seem to need some attention.
2. The `ESMC_FieldRegridDynMaskPredef*UTest.c` tests need to be reviewed and completed/corrected. They PASS in single -pet (UNI) mode but fail for multi-pet mode. There seem basic Grid setup problems that need to be addressed.
3. I have not yet reviewed the ESMPY layer. I think this only make sense _after_ the ESMF/ESMC Field layer issues have been addressed.

PR #578, opened 2026-08-03
Fix failures with AOCC compiler after Dynamic Mask merge

theurich:
This PR fixes failures that have been showing up in the nightlies with the APCC compiler after #523 has been merged.


oehmke (approved):
Looks good to me. Thanks for the fix.

PR #593, opened 2026-09-11
Preserve framework pairs in CMake exports

### LINKED ISSUE #592: Preserve macOS framework pairs in CMake package
## Summary

The generated `ESMFConfig.cmake` converts Apple framework pairs from ESMF link metadata into malformed CMake link-library entries.

## Reproduction

An ESMF build whose `ESMF_F90LINKLIBS` includes:

```text
-framework CoreFoundation -framework SystemConfiguration
```

generates a package that runs:

```cmake
separate_arguments(ESMF_TARGET_LINKS UNIX_COMMAND "${ESMF_LINK_LIBS_LIST}")
set_target_properties(ESMF::ESMF PROPERTIES
  INTERFACE_LINK_LIBRARIES "${ESMF_TARGET_LINKS}")
```

CMake then interprets the framework names as ordinary libraries and a consuming target receives:

```text
-framework -lCoreFoundation -lSystemConfiguration
```

The standalone `-framework` flag leaves the linker in framework-name mode, and the link fails, for example:

```text
ld: library 'SystemConfiguration' not found
```

This was reproduced with macOS, CMake 4.4.3, NAG 7.2.43, MPICH 5.0.1, and Baselibs 9.13.0 while linking MAPL. The issue is independent of the Fortran compiler.

## History

The `separate_arguments()` handling predates #580 in the generated-package template (introduced with #549). #580 correctly separated `ESMF_*LINKOPTS` into `INTERFACE_LINK_OPTIONS`, but framework pairs originate in `ESMF_*LINKLIBS`; it did not introduce this behavior.

## Proposed fix

When parsing link libraries in `cmake/FindESMF.cmake` and `cmake/ESMFConfig.cmake.in`, recognize each `-framework <name>` pair and publish it as one grouped interface link option, e.g. `SHELL:-framework CoreFoundation`. Remove those two tokens from `INTERFACE_LINK_LIBRARIES`. This preserves the required argument grouping despite CMake option de-duplication.

A MAPL-side workaround conditionally recognizes the broken exported sequence and substitutes `find_library()` framework paths, but ESMF should export a correct CMake target directly.

mathomp4:
## Summary

Preserve `-framework <name>` pairs from ESMF link metadata when exposing the CMake target interface.

CMake previously received each pair as independent `INTERFACE_LINK_LIBRARIES` items, transformed the framework name into `-l<name>`, and left a bare `-framework` flag. This produces link lines such as:

```text
-framework -lCoreFoundation -lSystemConfiguration
```

The change converts each pair to one `SHELL:-framework <name>` `INTERFACE_LINK_OPTIONS` item and removes its tokens from the library interface. It applies to both the generated config package and legacy `FindESMF.cmake` module.

## Validation

- `git diff --check`
- CMake 4.4.3 consumer fixture using generated config mode
- CMake 4.4.3 consumer fixture using legacy module mode and a Baselibs ESMF library
- Both fixtures linked successfully on macOS and emitted:

  ```text
  -framework CoreFoundation -framework SystemConfiguration
  ```

## Related issue

Fixes #592


mathomp4:
Note: GPT Terra helped out with this PR. It is...noisier than I expected.

For now, I do have a fix/workaround I can add to GEOS/MAPL's CMake that I'm testing.

mathomp4:
Closing for now. This might be due to Baselibs weirdness.

